PR #4, opened 2025-01-23
SYS-1732: SpaCy NLP - director name extraction proof of concept

ztucker4:
Implements [SYS-1732](https://uclalibrary.atlassian.net/browse/SYS-1732)

Adds a new script, `spacy_experiment.py`, demonstrating the use of spaCy to extract names for pre-extracted 245 $c data. The "small" and "medium" versions of spaCy's model are used. The script assumes the presence of the file `f245c_directors.txt`.

Since spaCy does not work with Python 3.13 (see discussion [here](https://github.com/explosion/spaCy/issues/13658)), the Dockerfile has been updated to use a base Python image with Python 3.12. While the latest version of spaCy is 3.8.4, `requirements.txt` specifies version 3.7.5 due to incompatibilities with Mac M1 architecture in later versions (discussion [here](https://github.com/explosion/cython-blis/issues/117#issuecomment-2596810409)). 

To test, rebuild the container (`docker compose build`) and run the script (`docker compose run ftva_data python spacy_experiment.py`). You should see terminal output describing the total number of entities. Two output files should be produced, `output_core_web_sm.csv` and `output_core_web_md.csv`. These are CSV files with three columns: the original 245 data, the names extracted by spaCy, and the non-name "entities" extracted by spaCy. 

[SYS-1732]: https://uclalibrary.atlassian.net/browse/SYS-1732?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


spacy_experiment.py
```diff
@@ -0,0 +1,83 @@
...
+    print(f"output written to {file_name}")
+
+
+def main():
+    with open("f245c_directors.txt", encoding="utf-8") as file:
+        raw_data = file.read()
+
+    data = ast.literal_eval(raw_data)
+    # concatenate each subfield_c into a single string
+    for i, subfield_c in enumerate(data):
```
akohler, line 69: This can be simpler via list comprehension:
`data = [" ".join(row) for row in data]`

akohler (changes requested):
@ztucker4 Thanks for putting this together.  Neither the Small nor Medium model is perfect (nor the Large, which I also loaded and tested), but overall Medium is best with this data:
```
Differences between Small and Medium: 27

Model: Small
Better than Medium: 8
Left word out of name: 7
Left name out completely: 11
Tokenized name wrong: 1

Model: Medium
Better than Small: 17
Left word out of name: 5
Left name out completely: 3
Tokenized name wrong: 2
```

A few small changes before merging: switching from `ast` to `json` (since we know the source of the sample data), and a more Pythonic way to flatten the list of lists into a list of strings.
Thanks --Andy

akohler (approved):

PR #9, opened 2025-02-13
SYS-1740: script to identify monographs given Alma and FM data

ztucker4:
Implements [SYS-1740](https://uclalibrary.atlassian.net/browse/SYS-1740)

A new script to identify monographs using the criteria in the [A1 Monographs](https://uclalibrary.atlassian.net/wiki/spaces/pm/pages/909410326/A1+Monographs) requirements document. The script uses Form (a combination of MARC 655 and FM `production_type`), plus Title (MARC 245 $n,$p, 250, 505), to determine which input records are monographs. Monograph bib records are output as a MARC file.

This version of the script is intended to align with version 9 of the [A1 Monographs doc](https://uclalibrary.atlassian.net/wiki/spaces/pm/pages/909410326/A1+Monographs?pageVersion=9) and version 3 of the [Form Terms Transformations](https://uclalibrary.atlassian.net/wiki/spaces/pm/pages/926220312/Form+Terms+Transformations) doc.

Inputs:

- `--alma_holdings_file`: path to a JSON file, assumed to be formatted like `alma_match_data.json` as attached to [SYS-1727](https://uclalibrary.atlassian.net/browse/SYS-1727). Using this input data allows us to avoid going back to the Alma API to get the Call Number that we need to match with FileMaker data.
- `--alma_bibs_file`: path to a file containing MARC bib records, as produced by `get_alma_bib_records.py`. 
- `--filemaker_data_file`: path to a CSV file as exported by FileMaker, containing precisely `inventory_no` and `production_type` values (without headers, as is default for FM exports). Let me know if you would like to send over the file I used, or you should be able to export this yourself from the `labeling` database in seconds.
- `--output_file`: desired path to the output file containing full MARC bib records of monographs only.
- `--form_mismatch_report`: optional, desired path to a CSV file reporting on differences between FM `production_type` and Alma 655 data post-normalization.

In addition to the MRC output and optional mismatch report, the script will output a log file, `identify_monographs_{timestamp}.log`. 

Using the `alma_match_data.json` file attached to [SYS-1727](https://uclalibrary.atlassian.net/browse/SYS-1727), a fresh bibs file produced by `get_alma_bib_records.py` using `alma_match_data.json`, and a recent export of FM data, I see 802 monographs identified out of 1777 records. 434 of these monographs have Form data mismatches.

Example usage: 
```docker compose run ftva_data python identify_monographs.py --alma_holdings_file alma_match_data.json --alma_bibs_file alma_match_data_bibs.mrc --filemaker_data_file inventory_no_production_type.csv --output_file monographs.mrc --form_mismatch_report mismatch.csv```

[SYS-1740]: https://uclalibrary.atlassian.net/browse/SYS-1740?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SYS-1727]: https://uclalibrary.atlassian.net/browse/SYS-1727?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SYS-1727]: https://uclalibrary.atlassian.net/browse/SYS-1727?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/identify_monographs.py
+++ b/identify_monographs.py
@@ -14,27 +14,27 @@ def _get_args() -> argparse.Namespace:
     parser = argparse.ArgumentParser()
     parser.add_argument(
         "--alma_holdings_file",
-        help="Path to input file containing Alma MMS and Holdings data",
+        help="Path to input file containing Alma MMS and Holdings data (JSON)",
         required=True,
     )
     parser.add_argument(
         "--alma_bibs_file",
-        help="Path to input file containing Alma bib records",
+        help="Path to input file containing Alma bib records (MARC)",
         required=True,
     )
     parser.add_argument(
         "--filemaker_data_file",
-        help="Path to input file containing FileMaker data",
+        help="Path to input file containing FileMaker data (CSV)",
         required=True,
     )
     parser.add_argument(
         "--output_file",
-        help="Path to output file of monograph records only",
+        help="Path to output file of monograph records only (MARC)",
         required=True,
     )
     parser.add_argument(
         "--form_mismatch_report",
-        help="Optional: Path to output file of form mismatch report",
+        help="Optional: Path to output file of form mismatch report (CSV)",
         required=False,
     )
     args = parser.parse_args()
--- a/identify_monographs.py
+++ b/identify_monographs.py
@@ -1,0 +1,402 @@
+import argparse
+from get_ftva_alma_data import _get_subfields
+import json
+import csv
+import pymarc
+import logging
+from datetime import datetime
+from pathlib import Path
+from pymarc import Record
+
+
+def _get_args() -> argparse.Namespace:
+    """Returns the command-line arguments for this program."""
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "--alma_holdings_file",
+        help="Path to input file containing Alma MMS and Holdings data",
+        required=True,
+    )
+    parser.add_argument(
+        "--alma_bibs_file",
+        help="Path to input file containing Alma bib records",
+        required=True,
+    )
+    parser.add_argument(
+        "--filemaker_data_file",
+        help="Path to input file containing FileMaker data",
+        required=True,
+    )
+    parser.add_argument(
+        "--output_file",
+        help="Path to output file of monograph records only",
+        required=True,
+    )
+    parser.add_argument(
+        "--form_mismatch_report",
+        help="Optional: Path to output file of form mismatch report",
+        required=False,
+    )
+    args = parser.parse_args()
+    return args
+
+
+def _get_logger(name: str | None = None) -> logging.Logger:
+    """Returns a logger for this program."""
+    if not name:
+        # Use base filename of current script.
+        name = Path(__file__).stem
+    timestamp = datetime.now().strftime("%Y%m%d_%H%M%S")
+    logging_filename = f"{name}_{timestamp}.log"
+    logger = logging.getLogger(name)
+    logging.basicConfig(
+        filename=logging_filename,
+        level=logging.DEBUG,
+        format="%(asctime)s %(levelname)s: %(message)s",
+    )
+    return logger
+
+
+def _has_fm_production_type(fm_record) -> bool:
+    """Given a FileMaker record, checks whether the record has a production type value."""
+    production_type = fm_record.get("production_type")
+    if production_type:
+        return True
+    return False
+
+
+def _has_non_monograph_fm_production_type(fm_record) -> bool:
+    """
+    Given a FileMaker record, checks whether the record has a production type value
+    indicating a non-monograph type. Returns True if it does, False if it does not.
+    """
+    non_monograph_types_lowercase = [
+        "television series",
+        "compilation",
+        "mini-series",
+        "news",
+        "newsreel",
+        "serials",
+    ]
+    production_types = fm_record["production_type"]
+    # check if any of the non-monograph types, in any case, are in the production types string
+    if any(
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: identify_monographs.py [hunk 1]
```diff
@@ -0,0 +1,402 @@
...
+
+def _has_non_monograph_alma_title(marc_record) -> bool:
+    """Given a MARC record, checks whether the record has a title value indicating
+    a non-monograph type."""
+    # if $p or $n is in the title, it's not a monograph
+    marc_245 = marc_record.get_fields("245")
+    for field in marc_245:
+        if field["p"] or field["n"]:
+            return True
+    # if any 250 fields are present, it's not a monograph
```
akohler, line 260: This could be simplified: `get_fields(*args)` returns a list of all fields matching any of `*args`, so just checking `get_fields("250", "505")` is enough.

identify_monographs.py
```diff
@@ -0,0 +1,402 @@
...
+    alma_bib_dict = {record["001"].data: record for record in alma_bib_data}
+    full_data = []
+    for alma_data, fm_data in matched_data:
+        alma_record = alma_bib_dict.get(alma_data["MMS Id"])
+        full_data.append((alma_record, fm_data))
+    return full_data
+
+
+def is_monograph(alma_record: Record, fm_record: dict) -> bool:
+    """Determines whether a given record is a monograph."""
```
akohler, line 295: This docstring should be clearer, since the method checks two kinds of "record", and the Alma record is only relevant if the FM record has no production type. 

I think the implementation could be clearer too. The two top-level tests are separate; could this be rewritten as an `if-else`, rather than two (unrelated?) `if`s? Maybe additional helper method(s) which return the monograph-ness of Alma record and FM record, individually? Then this `is_monograph` method is simpler, with Alma-specific and FM-specific logic in those separate helper methods.

identify_monographs.py
```diff
@@ -0,0 +1,402 @@
...
+    """
+    forms = {}
+    if _has_fm_production_type(fm_record):
+        production_types = _get_fm_monograph_production_types(fm_record)
+        normalized_production_types = _normalize_fm_monograph_production_types(
+            production_types
+        )
+        forms["filemaker"] = normalized_production_types
+    monograph_forms = _get_alma_monograph_forms(alma_record)
+    normalized_monograph_forms = _normalize_alma_monograph_form_values(monograph_forms)
```
akohler, line 324: Since we only want to work with the normalized forms, I would move the call to `_normalize_alma_monograph_form_values` into `_get_alma_monograph_forms`, so that it normalizes the list before returning it.

identify_monographs.py
```diff
@@ -0,0 +1,402 @@
...
+def _write_monograph_records(records: list, output_file: str) -> None:
+    """Writes monograph records to a MARC file."""
+    with open(output_file, "wb") as f:
+        for alma_record, fm_record in records:
+            f.write(alma_record.as_marc())
+
+
+def main():
+    args = _get_args()
+    alma_data_refs = json.load(open(args.alma_holdings_file))
```
akohler, line 377: I would move this logic to a helper method. That has the minor benefit of allowing the `open` to be done via context manager, so the file gets closed, but also just hides this detail away from the main method.

identify_monographs.py
```diff
@@ -0,0 +1,402 @@
...
+    fm_data = csv.DictReader(
```
akohler, line 378: Similar to the alma data refs, I'd move the two commands which get `fm_data` to a helper method. This hides the details of opening file, field names, and list building, none of which needs to be in the main method.

identify_monographs.py
```diff
@@ -0,0 +1,402 @@
...
+    full_matched_records = _get_full_bib_data(matched_records, args.alma_bibs_file)
+    print(f"Matched {len(full_matched_records)} full records")
+    monograph_records = []
+    for alma_record, fm_record in full_matched_records:
+        if is_monograph(alma_record, fm_record):
+            monograph_records.append((alma_record, fm_record))
+    print(f"Found {len(monograph_records)} monograph records")
+
+    if args.form_mismatch_report:
+        print(f"Writing form mismatch report to {args.form_mismatch_report}")
```
akohler, line 393: I think print statements like this belong in the method they're associated with, since they're so closely related.

identify_monographs.py
```diff
@@ -0,0 +1,402 @@
...
+        _write_form_mismatch_report(monograph_records, args.form_mismatch_report)
+
+    print(f"Writing monograph records to {args.output_file}")
```
akohler, line 396: Move print statement to the associated method.

akohler (changes requested):
@ztucker4 The program runs very quickly, and I got the same record counts as you.

Several minor changes, mostly structural or organizational.  I also think the `print` statements probably should be using the `logger` instead, though it is convenient to get high-level summary output like this instead of buried in an otherwise-detailed log, so either way is OK with me.

Thanks --Andy

ztucker4:
Ready for re-review. Type hints and docstrings are fixed throughout. New helper functions are added for loading data, and for checking all Alma data for non-monographness (to simplify `is_monograph()`. For now, I'd like to keep the print statements as is for easier visibility. Thanks!

akohler (approved):

PR #10, opened 2025-02-25
SYS-1743: Add sample JSON documents, manually created for now

akohler:
@ztucker4 Here are the sample JSON documents we've started working on with Thelma.  I think the repository is the best place for these.

No real review needed, but please take a look and merge, then use the repo as the official source of truth as you all work on these.

Thanks --Andy


ztucker4 (approved):
Thanks Andy! JSON files are (now :wink:) correctly formatted, and it's nice to have these together in one place. Merging!

PR #13, opened 2025-04-04
SYS-1783: Add script to extract inventory numbers from FTVA spreadsheet

henry-r18:
Implements [SYS-1783](https://uclalibrary.atlassian.net/browse/SYS-1783)

**Summary**
- Loads data from FTVA digital lab spreadsheet
- Matches values against inventory number pattern described by FTVA staff
  - Only outputs first match according to specified column priority
  - If multiple matches in a single input string, outputs pipe-delimited list
- Writes new Excel spreadsheet with extracted inventory numbers in the `Inventory #` column

**Testing**
I used [regex101](https://regex101.com/) to assemble and test the regular expression I used. I also wrote a short test for the regular expression using `pytest` locally, but since `pytest` isn't an existing requirement for the dev container, I didn't bother including that test in this PR.

**Running script**
From within the dev container, run: `python extract_inventory_numbers.py --data_file ${PATH_TO_DL_SPREADSHEET}`

**Note**
You may need to install the `openpyxl` package, as `pandas` uses it for reading and writing to Excel formats.

[SYS-1783]: https://uclalibrary.atlassian.net/browse/SYS-1783?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/extract_inventory_numbers.py
+++ b/extract_inventory_numbers.py
@@ -19,10 +19,16 @@ def _get_args() -> argparse.Namespace:
 def _compile_regex() -> re.Pattern:
     # Return a compiled RegEx pattern for matching inventory numbers
     # the RegEx is comprised of three main parts, which are explained inline below
+    #
+    # NOTE: the pattern works in most cases, but will return false positives
+    # where the input contains a substring that syntactically matches pattern
+    # but is not actually a valid inventory number, according to FTVA
+    # e.g. AAC442\Title_T01ASYNC_Surround matches T01, which is a valid pattern,
+    # but not an actual inventory number
     regex_components = [
         r'(?<![A-Z])',  # pattern not preceded by capital letter, to mitigate false positives
         r'(?:M|T|DVD|FE|HFA|VA|XFE|XFF|XVE)',  # non-capturing group of 8 prefixes defined by FTVA
-        r'\d+',  # 1 or more digits, as many as possible
+        r'\d{2,}',  # 2 or more digits, as many as possible
         r'(?:[A-Z](?![A-Za-z]))?'  # optional suffix 1 capital letter not followed by another letter
     ]
 
@@ -41,7 +47,10 @@ def _extract_inventory_numbers(
     if matches:
         # per FTVA spec, provide pipe-delimited string
         # if multiple matches in a single input value
-        return '|'.join(set(matches))  # using set() to provide only unique inv #s
+        # uses dict.fromkeys() to get unique values
+        # while maintaining list order
+        return '|'.join(list(dict.fromkeys(matches)))
+    return ''
 
 
 def main() -> None:
--- a/extract_inventory_numbers.py
+++ b/extract_inventory_numbers.py
@@ -3,10 +3,6 @@ import re
 import pandas as pd
 from pathlib import Path
 
-# Regular expression to match inventory numbers,
-# per FTVA specification
-INVENTORY_NUMBER_PATTERN = re.compile(r'(?:M|T|DVD|HFA|VA|XFE|XFF|XVE)\d+(?:[A-Z](?![A-Za-z]))?')
-
 
 def _get_args() -> argparse.Namespace:
     """Returns the command-line arguments for this program."""
@@ -20,22 +16,49 @@ def _get_args() -> argparse.Namespace:
     return args
 
 
+def _compile_regex() -> re.Pattern:
+    # Return a compiled RegEx pattern for matching inventory numbers
+    # the RegEx is comprised of three main parts, which are explained inline below
+    regex_components = [
+        r'(?:M|T|DVD|HFA|VA|XFE|XFF|XVE)',  # non-capturing group of 8 prefixes defined by FTVA
+        r'\d+',  # 1 or more digits, as many as possible
+        r'(?:[A-Z](?![A-Za-z]))?'  # optional suffix 1 capital letter not followed by another letter
+    ]
+
+    return re.compile(''.join(regex_components))
+
+
 def _extract_inventory_numbers(df: pd.DataFrame) -> pd.DataFrame:
+
+    inventory_number_pattern = _compile_regex()
+
     # Priority in which to consider columns,
     # per instructions from FTVA
     column_priority = ('Source Inventory #', 'File Name', 'Sub Folder', 'File Folder Name')
+    existing_inventory_number_values = []
 
     for index, row in df.iterrows():
+        # The header row in the source spreadsheet is repeated many times
+        # as if multiple sheets were appended together
+        # this condition checks if the current value in the Inventory # column is:
+        # 1) not a header, and;
+        # 2) not NaN
+        # meaning it is some other value
+        # row indexes for any such values are then appended to a list to be printed for reference
+        if (not row['Inventory #'] == 'Inventory #') and (not pd.isna(row['Inventory #'])):
+            existing_inventory_number_values.append(index)
+
         for column in column_priority:
             if not isinstance(row[column], str):
                 continue
 
-            matches = re.findall(INVENTORY_NUMBER_PATTERN, row[column])
+            matches = re.findall(inventory_number_pattern, row[column])
             if matches:
                 # per FTVA spec, provide pipe-delimited string
                 # if multiple matches in a single input value
                 row['Inventory #'] = '|'.join(matches)
                 break  # we only want the first match in a row
+    print(df.iloc[existing_inventory_number_values]['Inventory #'])
     return df
 
 
--- a/extract_inventory_numbers.py
+++ b/extract_inventory_numbers.py
@@ -20,7 +20,8 @@ def _compile_regex() -> re.Pattern:
     # Return a compiled RegEx pattern for matching inventory numbers
     # the RegEx is comprised of three main parts, which are explained inline below
     regex_components = [
-        r'(?:M|T|DVD|HFA|VA|XFE|XFF|XVE)',  # non-capturing group of 8 prefixes defined by FTVA
+        r'(?<![A-Z])',  # pattern not preceded by capital letter, to mitigate false positives
+        r'(?:M|T|DVD|FE|HFA|VA|XFE|XFF|XVE)',  # non-capturing group of 8 prefixes defined by FTVA
         r'\d+',  # 1 or more digits, as many as possible
         r'(?:[A-Z](?![A-Za-z]))?'  # optional suffix 1 capital letter not followed by another letter
     ]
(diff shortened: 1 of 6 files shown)


extract_inventory_numbers.py
```diff
@@ -0,0 +1,59 @@
+import argparse
+import re
+import pandas as pd
```
akohler, line 3: Since `openpxyl` is a dependency now, please add it to `requirements.txt`.  Since `pandas` is such a big package, it doesn't automatically get installed with everything, and even installing `pandas[excel]` is overkill.  Check out the [detailed documentation](https://pandas.pydata.org/pandas-docs/stable/getting_started/install.html#excel-files) for more.

For this project, looks like I added `pandas` just for Salik's `excel2xml.py`, one of several initial scripts he provided but which we haven't really used, so we didn't catch the Excel dependency issue until now. I generally work with just CSV/TSV files, but exporting the whole Google document as Excel is convenient, so it's good to bring in this dependency now.

extract_inventory_numbers.py
```diff
@@ -0,0 +1,59 @@
...
+from pathlib import Path
+
+# Regular expression to match inventory numbers,
+# per FTVA specification
+INVENTORY_NUMBER_PATTERN = re.compile(r'(?:M|T|DVD|HFA|VA|XFE|XFF|XVE)\d+(?:[A-Z](?![A-Za-z]))?')
```
akohler, line 8: This isn't used globally by the script, so doesn't need to be a global constant.  It can just be declared in `_extract_inventory_numbers()` where it's used.

akohler, line 8: Also, please add a comment documenting the regular expression in general terms.

akohler (changes requested):
@henry-r18 This looks great. The flow and the code are clear, other than the regex which should be moved to the relevant method and explained via comments.

Please do add tests.  These should be written without requiring `pytest` as a dependency, though I'm happy to hear arguments why `pytest` improves things enough over the built-in testing framework to be worth considering :)

Tests should focus on proving the regex correctly handles all good cases, and correctly handles all bad cases the specs describe and any others you can anticipate.

Thanks --Andy

extract_inventory_numbers.py
henry-r18: Added `openpyxl` to `requirements.txt`

extract_inventory_numbers.py
henry-r18: I moved the regex pattern and compile to its own method, to address this comment and to make it easier to import into the test module.

extract_inventory_numbers.py
```diff
@@ -0,0 +1,82 @@
...
+            if not isinstance(row[column], str):
+                continue
+
+            matches = re.findall(inventory_number_pattern, row[column])
+            if matches:
+                # per FTVA spec, provide pipe-delimited string
+                # if multiple matches in a single input value
+                row['Inventory #'] = '|'.join(matches)
+                break  # we only want the first match in a row
+    print(df.iloc[existing_inventory_number_values]['Inventory #'])
```
akohler, line 61: Please print a message before this, explaining what the output is.

tests/test_extract_inventory_numbers.py
```diff
@@ -0,0 +1,31 @@
...
+from extract_inventory_numbers import _compile_regex
+
+
+class TestRegEx(unittest.TestCase):
+
+    def test_regex(self):
+        # test cases comprised of tuples
+        # where the first item is the input string
+        # and the second item is the expected output derived from FTVA specs
+        test_cases = (
```
akohler, line 12: Please add test cases for `T` and `DVD`.  Yes, they should work the same as all the others prefixes, but let the tests prove that.

tests/test_extract_inventory_numbers.py
```diff
@@ -0,0 +1,31 @@
...
+            ("XFE1915MX", "XFE1915"),  # invalid suffix case
+            ("Randy_Requiem1", ""),
+            (
+                "XFE4098M_XFF104M_DamagedLives_Finals",
+                "XFE4098M|XFF104M"
+            ),  # multi-match to pipe-delimited string case
+            (
+                "XVE779T_ZVE780T_OneNightStand_WorldOfLennyBruce_CaptureFiles_SD_2997FPS_YUV",
+                "XVE779T"
+            )  # close-but-no-cigar case (ZVE780T does not have valid prefix)
```
akohler, line 25: Please tweak this comment to make it clearer what's being tested: 1 good prefix, 1 bad.  I had to read it twice to be sure it wasn't a `XVE` vs `ZVE` typo.

tests/test_extract_inventory_numbers.py
```diff
@@ -0,0 +1,31 @@
...
+    def test_regex(self):
+        # test cases comprised of tuples
+        # where the first item is the input string
+        # and the second item is the expected output derived from FTVA specs
+        test_cases = (
+            ("HFA27M_Reel", "HFA27M"),
+            ("VA13161T_KTLA", "VA13161T"),
+            ("M190816Medea2", "M190816"),
+            ("XFE1915MX", "XFE1915"),  # invalid suffix case
+            ("Randy_Requiem1", ""),
```
akohler, line 17: Add a comment to document that this is an invalid prefix.

tests/test_extract_inventory_numbers.py
```diff
@@ -0,0 +1,31 @@
...
+            (
+                "XFE4098M_XFF104M_DamagedLives_Finals",
+                "XFE4098M|XFF104M"
+            ),  # multi-match to pipe-delimited string case
+            (
+                "XVE779T_ZVE780T_OneNightStand_WorldOfLennyBruce_CaptureFiles_SD_2997FPS_YUV",
+                "XVE779T"
+            )  # close-but-no-cigar case (ZVE780T does not have valid prefix)
+        )
```
akohler, line 26: Please add test case(s) where there's a possible valid prefix, but no valid inventory number.  For example:
* row 16360: `Max_From_DVD_H264`
* row 1227: `HARVEST3000AUDIOMAG` (I think this is invalid, but might match specs... not sure)
* row 1757: `M46293SanFernandoCLEANEXPORT3` (the `T3` might get included, though not valid - maybe another false positive)

If those last 2 are false positives, ask Thelma if there's any practical way to improve the specs.  For example, minimum number of digits?

akohler (changes requested):
@henry-r18 This is almost ready.  A couple of minor comment-related tweaks, and a few more test cases, especially looking for edge cases, then I'll merge this.
Thanks --Andy

tests/test_extract_inventory_numbers.py
henry-r18: @akohler, those last two cases (`HARVEST3000AUDIOMAG` and `M46293SanFernandoCLEANEXPORT3`) did indeed break my RegEx, as `T3000` substring matches the pattern in the first case and `T3` in the latter. Thanks for finding those!

I'll ask Thelma if there's a way to restrict the pattern further.

extract_inventory_numbers.py
henry-r18, line 1: The additional regex clause, which is the first in the list of regex components, should address the edge case where capitalized words with ending letters that happen to be part of the prefix list and are followed by digits get matched by the pattern.

extract_inventory_numbers.py
```diff
@@ -20,46 +20,28 @@ def _compile_regex() -> re.Pattern:
...
         r'(?:[A-Z](?![A-Za-z]))?'  # optional suffix 1 capital letter not followed by another letter
     ]
 
     return re.compile(''.join(regex_components))
 
 
-def _extract_inventory_numbers(df: pd.DataFrame) -> pd.DataFrame:
+def _extract_inventory_numbers(
+        value: str,
+        inventory_number_pattern: re.Pattern = _compile_regex()
```
henry-r18, line 34: I moved the pattern to be a default parameter for the function, to make the function more adaptable.

extract_inventory_numbers.py
```diff
@@ -20,46 +20,28 @@ def _compile_regex() -> re.Pattern:
...
+        r'(?:M|T|DVD|FE|HFA|VA|XFE|XFF|XVE)',  # non-capturing group of 8 prefixes defined by FTVA
         r'\d+',  # 1 or more digits, as many as possible
         r'(?:[A-Z](?![A-Za-z]))?'  # optional suffix 1 capital letter not followed by another letter
     ]
 
     return re.compile(''.join(regex_components))
 
 
-def _extract_inventory_numbers(df: pd.DataFrame) -> pd.DataFrame:
+def _extract_inventory_numbers(
```
henry-r18, line 32: I revised this function quite a bit, to better suit the new extract logic.

tests/test_extract_inventory_numbers.py
henry-r18, line 1: I improved the inline comments on the test cases, as well as adding the cases you identified @akohler

extract_inventory_numbers.py
```diff
@@ -0,0 +1,68 @@
...
+    )
+    args = parser.parse_args()
+    return args
+
+
+def _compile_regex() -> re.Pattern:
+    # Return a compiled RegEx pattern for matching inventory numbers
+    # the RegEx is comprised of three main parts, which are explained inline below
+    regex_components = [
+        r'(?<![A-Z])',  # pattern not preceded by capital letter, to mitigate false positives
```
henry-r18, line 23: This additional regex clause addresses the edge case where capitalized words with ending letters that happen to be part of the prefix list and are followed by digits get matched by the pattern.

henry-r18:
@akohler 

I confirmed with Thelma on Slack that valid inventory numbers should include 2 or more digits, so the `T1...T4` cases you identified should not match. I'm not sure how to create a regex clause that ignores the last case `T01` without explicitly excluding that specific substring. Since it appears to be the only case in the current data, I did not adjust the regex to account for it, but rather made a note that it's a known false positive.

**Latest commit**

I pushed another commit that includes the relevant change to the regex pattern, as well as:

- fixes an issue where the matches were returning out of their original order, because I used `set()` to drop duplicates
- fixes an issue where the `_extract_inventory_numbers` function returned `None` instead of an empty string
- adds tests for the new edge cases discussed above
- revised the test to target `_extract_inventory_numbers`, rather than _compile_regex, now that `_extract_inventory_numbers` is more modular

**Rebase from `main`**

I used `git pull --rebase` on my local feature branch to pull in Zoe's recent commit from `main`, but I'm not sure I did it right. Perhaps we can look at it together at stand up.

Thanks!
Henry

henry-r18:
@akohler 

It looks to me like the regex is working as expected on `AAC442\Title_T01ASYNC_Surround`, because the last clause is meant to capture the optional suffix, which this input doesn't have. If the input was `T01A_SYNC`, the pattern would match the optional suffix, which has to be a single capital letter not followed by any other letter. But in the case at issue, we'd still want to capture T01, as it has a valid prefix and a valid number of digits, just no optional suffix. It just so happens that this inventory number isn't real, because Thelma says so.

My solution was to hard-code a list of "known false positives" (i.e. things that look like inventory numbers but aren't actually), then remove them from the matched list before returning it, as you suggested. I updated the relevant test case accordingly, and all the tests pass.

I also added to the README, which then resulted in a merge conflict. I attempted to resolve that merge conflict in yet another commit. I think all is well at this point...

-Henry

PS Thanks for walking me through git rebasing earlier—things are much clearer for me now!

akohler (approved):
@henry-r18 This all looks good to me now.  Your "known false positives" solution is fine, given the limited number (only 1 so far) and the messiness of the source data.  I re-checked the output and confirmed `T01` is no longer included.  All tests pass, test cases are meaningful and look complete to me.

Merging now, thanks.  @ztucker4 FYI.

PR #18, opened 2025-07-16
SYS-1913: Create metadata processing script, including Creator values

ztucker4:
Implements [SYS-1913](https://uclalibrary.atlassian.net/browse/SYS-1913)

Adds a new script, `generate_metadata.py`, intended for use with 1-1-1 matches. The script takes a CSV file containing Alma MMS IDs (plus other identifiers) and creates a JSON file with the specified filename. For now, the JSON file contains one object for each row. Each object contains the data from the original CSV plus extracted Creator names.

Script arguments: 
* `--config_file`: a TOML config file, as used by other scripts in this repo. This is currently only used for the `alma_api_key` value.
* `--input_file`: a CSV file containing identifiers. Expected column headers are `alma_bib_id`, `alma_holdings_id`, `pd_record_id`, and `django_record_id`.
* `--output_file` (optional): desired path to the output JSON file. Defaults to `processed_metadata.json`.

Notes:

* The [Test Plan Confluence page](https://uclalibrary.atlassian.net/wiki/spaces/pm/pages/1293058248/Test+Plan+-+July+2025) does not include MMS IDs. We need these in order to get bibliographic metadata from Alma, and the Alma API doesn't seem to offer any way to work backwards and identify a MMS ID associated with a given Holdings ID. For now, I'm assuming we can include MMS IDs in input data.
* The [Alma record](https://search.library.ucla.edu/permalink/01UCS_LAL/17p22dp/alma991276403506533) indicated in the Test Plan does not have Creator information (245 $c). So to test creator extraction, I grabbed Alma IDs for a different Twilight Zone episode with a 245 $c containing a director and included it as the second row in the CSV data below.

No automated tests are included with this PR. To try the script manually, save the below as `sample_input.csv`, then run (in the container):
```python generate_metadata.py --input_file sample_input.csv --config_file prod_config_secrets.toml```. The output JSON file will be created as `processed_metadata.json`. A short log file will be created as `generate_metadata_{date}_{time}.log`.

```
alma_bib_id,alma_holdings_id,pd_record_id,django_record_id
991276403506533,22549193640006533,408716,999
991702153506533,22548952040006533,123456,888
```

[SYS-1913]: https://uclalibrary.atlassian.net/browse/SYS-1913?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/generate_metadata.py
+++ b/generate_metadata.py
@@ -1,0 +1,218 @@
+import csv
+import json
+import argparse
+import tomllib
+import spacy
+import logging
+from datetime import datetime
+from pathlib import Path
+from pymarc import Record
+from alma_api_client import AlmaAPIClient, get_pymarc_record_from_bib
+
+# for type hinting
+from spacy.language import Language
+
+
+def _get_arguments() -> argparse.Namespace:
+    """Parse command line arguments.
+
+    :return: Parsed arguments for config_file, input_file, and
+    output_file as a Namespace object."""
+    parser = argparse.ArgumentParser(description="Process metadata for MAMS ingestion.")
+    parser.add_argument(
+        "--config_file", help="Path to configuration file", required=True
+    )
+    parser.add_argument(
+        "--input_file",
+        type=str,
+        required=True,
+        help="Path to the input CSV file containing Alma holdings IDs and other metadata.",
+    )
+    parser.add_argument(
+        "--output_file",
+        type=str,
+        required=True,
+        help="Path to the output JSON file where processed metadata will be saved.",
+    )
+    return parser.parse_args()
+
+
+def _get_config(config_file_name: str) -> dict:
+    """Returns configuration for this program, loaded from TOML file.
+
+    :param config_file_name: Path to the configuration file.
+    :return: Configuration dictionary."""
+
+    with open(config_file_name, "rb") as f:
+        config = tomllib.load(f)
+    return config
+
+
+def _get_logger(name: str | None = None) -> logging.Logger:
+    """
+    Returns a logger for the current application.
+    A unique log filename is created using the current time, and log messages
+    will use the name in the 'logger' field.
+
+    :param name: Optional name for the logger. If not provided, uses the base filename
+    of the current script.
+    :return: Configured logger instance."""
+
+    if not name:
+        # Use base filename of current script.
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: generate_metadata.py [hunk 1]
```diff
@@ -0,0 +1,220 @@
...
+    for phrase in attribution_phrases:
+        if phrase in source_string.lower():
+            start_index = source_string.lower().find(phrase) + len(phrase)
+            # Extract substring after the phrase until the end of the string
+            # TODO: Does this work in general? Could a director be listed before
+            # a non-director role that we shoudn't include?
+            creator_string = source_string[start_index:].strip()
+            break
+
+    if not creator_string:
```
akohler, line 140: This will fail if nothing is found: pylance points out "creator_string is possibly unbound".  A simplified example:
```
from generate_metadata import _parse_creators
import spacy
nlp_model = spacy.load("en_core_web_md")

_parse_creators("film directed by John Smith", nlp_model)
# Success: ['John Smith']

_parse_creators("made up by John Smith", nlp_model)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/ftva_data/project/generate_metadata.py", line 140, in _parse_creators
    if not creator_string:
           ^^^^^^^^^^^^^^
UnboundLocalError: cannot access local variable 'creator_string' where it is not associated with a value
```

akohler (commented):
@ztucker4 Overall this is great.  Good catch on the bib id requirement; we'll  probably need to change `get_perfect_matches.py` to include that. I had forgotten that the Alma API forces holdings to be retrieved via bib, and not by themselves. 

Also interesting that there's no creator in their sample record; I'll ask if they want to consider switching records now, or asking Cataloging to add one if possible.

There are a couple of things flagged by pylance: one type hint mismatch, which is not a blocker but something I don't want to ignore unless there's no practical workaround, which this one has.  There's also a potentially un-initialized variable, which is a potential run-time error.

Once those are corrected, I'll merge.  Thanks --Andy

ztucker4:
Thanks for the feedback! This is ready for another look.

PR #19, opened 2025-07-17
SYS-1918: add date to generate_metadata.py

ztucker4:
Implements [SYS-1918](https://uclalibrary.atlassian.net/browse/SYS-1918)

This PR adds date extraction to `generate_metadata.py`. A date is extracted from the first 260 ## $c, parsed into a standard format (if possible), and included in the output JSON as `release_broadcast_date`.

In order to do anything beyond trivial parsing (i.e. when dates are already YYYY-MM-DD), I'm using `dateutil`, which is added to `requirements.txt`. 

Note that this implementation does not preserve the level of precision of the original data. If an incomplete date is input, the `dateutil` parser will fill in a default value for missing data, currently using values from the current timestamp (so "April 2023" will come out as "2023-04-17" today, since it's 7/17). In order to correctly handle these cases, I think we'll need to use regexes to determine the original level of precision. Our sample data does have a full date value, so this implementation should work for our immediate needs.

Rebuild the container to install the new library, and then run: `python generate_metadata.py --input_file sample_input.csv --config_file prod_config_secrets.toml`

### DIFF
--- a/generate_metadata.py
+++ b/generate_metadata.py
@@ -4,6 +4,7 @@ import argparse
 import tomllib
 import spacy
 import logging
+import dateutil.parser
 from datetime import datetime
 from pathlib import Path
 from pymarc import Record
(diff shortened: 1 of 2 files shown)


akohler (approved):
@ztucker4 All good, merged.
@henry-r18 FYI, to rebase onto your WIP.

Thanks --Andy

PR #20, opened 2025-07-17
SYS-1916: Generate map of LC language codes and names

akohler:
Implements [SYS-1916](https://uclalibrary.atlassian.net/browse/SYS-1916).

This PR adds a script, `create_language_map.py`:
```
python create_language_map.py [--store_full_data]
```

This creates or replaces `language_map.json`, a simple JSON mapping of language code to language name for all of the
languages supported in MARC bibliographic records, per the Library of Congress (LC).  This includes obsolete languages,
and the collective name for language families where multiple languages are grouped by LC under one code.  This mapping is needed to support the `language` metadata element for MAMS-related ETL.

If the optional `--store_full_data` flag is passed, the program will also create or replace `languages_full_data.json`,
which contains the full JSON data, explicitly identifying obsolete languages, alternate names of languages,
and all languages under collective names. This is not currently used but could be useful for future projects.

`README.md` has been updated with this documentation.

Output:
With the `--store_full_data` flag, the two output files should be similar to these:
```
      11696 Jul 17 13:05 language_map.json
    542065 Jul 17 13:05 languages_full_data.json
```

Tests:
I added 3 simple tests for the only method which I think warrants them, `_get_language_value()`.  Including existing tests, there now are 18, all should be passing:
`docker compose exec ftva_data python -m unittest`

@ztucker4 please review; @henry-r18 FYI.

[SYS-1916]: https://uclalibrary.atlassian.net/browse/SYS-1916?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,3 +189,17 @@ Read 605353 records from filemaker_data_20250624_182542.json
 FM data: 553605 rows
 Found 2177 perfect matches.
 ```
+
+### Create a "language map" file with data for all languages used by MARC bib records
+
+```
+python create_language_map.py [--store_full_data]
+```
+
+This creates or replaces `language_map.json`, a simple JSON mapping of language code to language name for all of the
+languages supported in MARC bibliographic records, per the Library of Congress (LC).  This includes obsolete languages,
+and the collective name for language families where multiple languages are grouped by LC under one code.
+
+If the optional `--store_full_data` flag is passed, the program will also create or replace `languages_full_data.json`,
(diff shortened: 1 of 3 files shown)


ztucker4 (approved):
This all looks great to me. Code structure is clear, with helpful comments. The generated files look good, and the tests pass and make sense. And even if LC goes offline, at least I now have a copy of this data :)
Merging now!

PR #27, opened 2025-07-24
SYS-1922: retrieve Alma data by call number

ztucker4:
Implements [SYS-1922](https://uclalibrary.atlassian.net/browse/SYS-1922)

For now, a stand-alone script that retrieves Alma data by call number using the SRU protocol. To run: `python get_alma_data_by_call_number.py --config_file {config_file.toml} --call_number M03016`

Like the corresponding Filemaker script, this script will check if the given call number matches a single Alma record, and print the [specified MARC data fields](https://uclalibrary.atlassian.net/wiki/spaces/pm/pages/1305214995/Digital+Data+Preview+fields) if so.

The output JSON is currently pretty ugly, and will need to be cleaned up before display to end users. I considered trying to convert the original XML to a `pymarc` Record and doing manipulations on that object, but I don't think that would actually save us much effort when putting together the display.

[SYS-1922]: https://uclalibrary.atlassian.net/browse/SYS-1922?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/get_alma_data_by_call_number.py
+++ b/get_alma_data_by_call_number.py
@@ -131,8 +131,6 @@ def main() -> None:
             f"Multiple ({number_of_records}) records found for call number "
             f"{args.call_number}"
         )
-        print("Full Alma data:")
-        print(json.dumps(alma_data, indent=4))
     else:
         print(f"Single record found for call number {args.call_number}")
         relevant_fields = get_relevant_fields_alma(alma_data)


get_alma_data_by_call_number.py
```diff
@@ -0,0 +1,133 @@
...
+
+    :param alma_data: The full Alma data dictionary, as obtained from the SRU API,
+        assumed to contain only one record.
+    :return: A list of relevant fields extracted from the Alma data.
+    """
+    # SearchRetrieveResponse is the top-level key in the Alma SRU response
+    if "searchRetrieveResponse" not in alma_data:
+        return []
+
+    # Assuming there is only one record in the response, this is the expected structure
```
akohler, line 82: How is the structure different if multiple records are returned?  Let's document this now while looking at the responses.  For possible examples of inventory numbers which are in multiple records, you can use the CSV generated by `get_ftva_holdings_report.py` to find values which occur multiple times.

akohler (changes requested):
@ztucker4 No code changes needed now, this works as expected for single-record returns and 0-record returns.  My only requested change is to add brief documentation comments on how the multiple-record data structure is different.

I agree with you that the conversion done by `xmldict` is not pretty.  If the MARCXML record embedded in the SRU response is really valid MARCXML, we could consider pulling it out of the XML response directly, or even re-converting the record itself back into MARCXML via [dicttoxml](https://pypi.org/project/dicttoxml/) (unsurprisingly).  But that's out of scope for now, as you noted, either way we'll need to do some work to get usable data for display (and for the metadata generation).

ztucker4:
Hi @akohler, I've added comments describing the structure of multiple-record Alma SRU responses. In testing this, I discovered that my original code didn't work for call numbers containing spaces. This has been fixed.

akohler (approved):
@ztucker4 Thanks for adding the multiple-record documentation. And for fixing the embedded space issue.  Merging now, thanks --Andy

@henry-r18 FYI

PR #28, opened 2025-08-25
SYS-1974: Use `ftva_etl` package

henry-r18:
Implements [SYS-1974](https://uclalibrary.atlassian.net/browse/SYS-1974)

### Acceptance criteria
- [x] The `generate_metadata.py` script is rewritten to use the new `ftva_etl` package
- [x] The script reads from the new [list of records](https://docs.google.com/spreadsheets/d/1LAbSnTm7YCvkJRrerJKReAvWKyrGKQYNkH-ZecN4FvM/edit?gid=0#gid=0) for the 25-50 pilot
- [x] Assets and track files are written separately
  - [x] The relationship between assets and tracks is obtained from the spreadsheet, via the “Audio track?” and “match_asset” columns

### Description
This PR reworks the `generate_metadata.py` module to use our `ftva_etl` package. The script accepts an input file in the form of a CSV export of the "DD Record IDs_Media to Ingest into MAMs" Google Sheet linked above and a TOML config file with API credentials. Using the "Audio Track?" column, the script separates tracks from assets and retrieves the UUID for match assets from the provided DD record ID. Lists of track and asset data are then written as separate JSON files to an output directory that can be configured via CLI arguments.

Utilizing the `ftva_etl` package required updating to Python 3.13 in the Dockerfile, and dependencies in `requirements.txt` that have been consolidated into `ftva_etl` were removed.

I made two small changes to the way the logger was previously handled:
1. made it write to a `logs` directory; and
2. declared it in the whole module scope, so that it is created during tests.

### Testing
I updated `test_generate_metadata.py` in such a way that our unit-test suite will still pass, but the tests are still WIPs. I experimented with the [unittest.mock](https://docs.python.org/3/library/unittest.mock.html) object library, so as to avoid calls to the actual APIs.

The tests are not so meaningful at this point; mostly just showing that the central `process_input_data` function returns lists for tracks and assets. I could set up sample records from the three data sources based on real records, then manipulate them to test different edge scenarios.

[SYS-1974]: https://uclalibrary.atlassian.net/browse/SYS-1974?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.12-slim-bookworm
+FROM python:3.13-slim-bookworm
 
 # Install git, required to install the alma_api_client package from github.
 RUN apt-get update && \
--- a/generate_metadata.py
+++ b/generate_metadata.py
@@ -11,6 +11,7 @@ from ftva_etl import (
     DigitalDataClient,
     get_mams_metadata,
 )
+from requests.exceptions import HTTPError
 
 
 def _get_arguments() -> argparse.Namespace:
@@ -69,7 +70,7 @@ def _get_logger(name: str | None = None) -> logging.Logger:
     logger = logging.getLogger(name)
     logging.basicConfig(
         filename=logging_file,
-        level=logging.DEBUG,
+        level=logging.INFO,
         format="%(asctime)s %(levelname)s: %(message)s",
     )
     return logger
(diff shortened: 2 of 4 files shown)


henry-r18:
@akohler Thanks for your notes! I improved the logging per your suggestions, revised the exception handling to be more specific, and added `xmltodict` back into the requirements. I chose to remove the tests I'd added using `unittest.mock`, but I left the test file as a placeholder. Ready for another look, when you have a chance.

akohler (approved):
@henry-r18 All looks great now, thanks for making the changes.  Confirmed all tests in the repo pass, log output is clean, output files are renamed and counts match the log output, error handling looks appropriate.  Merging now.
@ztucker4 FYI

PR #46, opened 2026-03-31
DLSR-44: Batch updates for FM Language field

ztucker4:
Implements [DLSR-44](https://uclalibrary.atlassian.net/browse/DLSR-244)

Specs: https://uclalibrary.atlassian.net/wiki/spaces/pm/pages/1734213643/language 

Adds functionality for batch Language normalization updates in FM. For a dry run (takes about 45 minutes for me): ```docker compose exec ftva_data python filemaker_batch_update.py -c prod_config_secrets.toml -f Language --dry_run```

Note the capital L in `Language` - for now, the script expects field names input on the command line to exactly match those in FM. 

New test cases are added for common Language normalization tasks, for a total of 33.

For @henry-r18's work: in addition to defining new `TRANSFORMER`s, this PR updates `_split_multivalue_field()` and `_rejoin_multivalue_field()` to use explicitly defined delimiters. These are managed in the variable `FIELD_DELIMITERS`. 

[DLSR-44]: https://uclalibrary.atlassian.net/browse/DLSR-44?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/filemaker_batch_update.py
+++ b/filemaker_batch_update.py
@@ -2,6 +2,8 @@ import tomllib
 import logging
 import argparse
 import re
+from string import capwords
+from strsimpy.normalized_levenshtein import NormalizedLevenshtein
 from collections.abc import Iterator
 from datetime import datetime
 from pathlib import Path
@@ -110,7 +112,7 @@ def _get_config(config_file_name: str) -> dict:
 # as the number of fields grows.
 # --------------------
 MAPPINGS = {
-    # NOTE: to smooth out casing inconsistencies,
+    # NOTE: for production_type values, to smooth out casing inconsistencies,
     # all values (except special cases) will be uppercased prior to mapping,
     # so keys need to be provided in uppercase here to ensure consistent mapping.
     # These mappings are intended to standardize variants of the same term into a controlled list,
@@ -126,7 +128,23 @@ MAPPINGS = {
         "FULL SILENT APERTURE 1.33:1": None,  # None means the value will be removed
         "SF": None,
         "SE": None,
-    }
+    },
+    "language": {
+        "Fr": "French",
+        "Ital": "Italian",
+        "Eng": "English",
+        "Portuguese for Brazil": "Portuguese",
+        "Unknown": "Undetermined",
+        "?": "Undetermined",
+        "": "Undetermined",  # TODO: does this capture NULLs?
+        "N/A": "No linguistic content",
+        "None": "No linguistic content",
+    },
+}
+
+FIELD_DELIMITERS = {
+    "production_type": "\r",
+    "language": ", ",
 }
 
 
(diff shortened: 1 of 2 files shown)


filemaker_batch_update.py
```diff
@@ -160,6 +177,75 @@ def _make_uppercase(value: str) -> str:
...
+    if closest_language is not None and closest_distance < 0.2:
+        if closest_language != value:
+            # Log the normalization decision, but only if a change is actually being made
+            logger.debug(
+                f"Normalized language spelling: {value!r} -> {closest_language!r} "
+                f"(distance={closest_distance:.2f})"
+            )
+        return closest_language
+    else:
+        if value is not None and value.strip() != "":
```
akohler: This test feels odd here.  If the `value` passed in *is* effectively empty, we just did a bunch of comparisons for nothing.  Also, if `value = None`, this will crash with an unhandled `TypeError`:
```
>>> from filemaker_batch_update import _normalize_language_spelling
>>> value = None
>>> s = _normalize_language_spelling(value)
Traceback (most recent call last):
  File "<python-input-2>", line 1, in <module>
    s = _normalize_language_spelling(value)
  File "/home/ftva_data/project/filemaker_batch_update.py", line 230, in _normalize_language_spelling
    distance = levenshtein.distance(value, language)
  File "/home/ftva_data/.local/lib/python3.13/site-packages/strsimpy/normalized_levenshtein.py", line 33, in distance
    raise TypeError("Argument s0 is NoneType.")
TypeError: Argument s0 is NoneType.
```

I suggest putting testing `value` at the start of the method and returning an appropriate value (probably `''`) immediately, only proceeding to the distance tests if there's a real value.

filemaker_batch_update.py
```diff
@@ -171,42 +257,62 @@ def _make_uppercase(value: str) -> str:
         lambda value: MAPPINGS["production_type"].get(
             value, value
         ),  # Apply the mapping defined above
-    ]
+    ],
+    "Language": [
+        _trim_whitespace,
+        _make_capitalized,
+        _normalize_language_spelling,
```
akohler: Should the spelling normalization be done after removing repeated words and possible intertitles? I think doing a distance comparison first will usually not be close enough to succeed.
```
# Successfully fix a typo
>>> _normalize_language_spelling("Russoan")
'Russian'

# Same typo, with intertitles
>>> _normalize_language_spelling("Russoan Intertitles")
'Russoan Intertitles'

# No typo but repeated word
>>> _normalize_language_spelling("Russian Russian")
'Russian Russian'
```

filemaker_batch_update.py
```diff
@@ -171,42 +257,62 @@ def _make_uppercase(value: str) -> str:
...
-def _split_multivalue_field(value: str) -> list[str]:
-    """Split multivalue field (delimited by "\r" or ";") into a list of values."""
-    # Normalize delimiters to "\r"
-    value = value.replace(";", "\r")
-    return value.split("\r")
+def _split_multivalue_field(value: str, delimiter: str) -> list[str]:
+    """Split a multi-value field into a list of values, based on the provided delimiter.
+    Also trims whitespace from each value and filters out any empty values."""
+    if delimiter == ", ":
+        # Normalize all possible delimiters to comma for language
```
akohler, line 280: The comment implies this delimiter normalization is for languages only.  If so, I suggest using an enumeration for delimiters and referring to the enum value here, making it really clear this is specific to Language values.  I'll add a separate comment where `FIELD_DELIMITERS` currently is declared.

filemaker_batch_update.py
```diff
@@ -126,7 +128,22 @@ def _get_config(config_file_name: str) -> dict:
...
+        "PORTUGUESE FOR BRAZIL": "Portuguese",
+        "UNKNOWN": "Undetermined",
+        "?": "Undetermined",
+        "": "Undetermined",
+        "N/A": "No linguistic content",
+        "NONE": "No linguistic content",
+    },
+}
+
+FIELD_DELIMITERS = {
```
akohler: As this grows, it may be better as an enumeration (specifically, a `StrEnum`), like this:
```
>>> from enum import StrEnum
>>> class FieldDelimiters(StrEnum):
...     production_type = "\r"
...     Language = ", "
...
```

Then values can be accessed in a more semantic, self-documenting way:
```
# Simulating how this is used in `_apply_transformers()`
# field_name is passed into the method.
>>> field_name = "production_type"

# Get the enum itself
>>> FieldDelimiters[field_name]
<FieldDelimiters.production_type: '\r'>
# Or just its value in a general way, more likely
>>> FieldDelimiters[field_name].value
'\r'

# Other places where field-specific meaning is more important
>>> FieldDelimiters.Language
<FieldDelimiters.Language: ', '>
>>> FieldDelimiters.Language.value
', '

# Code in `_split_multivalue_field()` could be:
if delimiter == FieldDelimiters.Language.value:
    ...
```

akohler (changes requested):
@ztucker4 I think this does a good job of extending the implementation to allow language updates.  As the number of fields grows, refactoring will become more important; this is probably as far as it should go with the current structure.

I did flag what I think are a couple of logic errors, where the sequence of transformations is probably making the distance check less effective than it should be.  The overall sequence does wind up working as it currently is, mainly because the data has repeated correct values instead of repeated typos, but the distance check should be applied after other cleanup is done.

There's one place where I think refactoring is worth doing now (pre-factoring?). Turning the current `FIELD_DELIMITERS` dictionary into an enumeration could make the usage of those values clearer overall, and should be just as easy to maintain / extend as new fields are added.

Thanks --Andy

ztucker4:
Hi @akohler, this is ready for another look. Field delimiters are now contained in a `StrEnum` object, and I updated the logic for None/empty string values in `_normalize_language_spelling`. 

I also moved `_normalize_language_spelling` to the second-to-last step in the transformers, just before the final mapping to known values. In a smallish sample of FM records I didn't see any values that would benefit from the re-ordering, but it definitely makes more sense to do the intertitles and deduplication steps first.

akohler (approved):
@ztucker4 Thanks for making the changes; seeing them in context, I think they do make the code more readable.

Tests all still pass.  I did a dry run on the first 500 records to see what the logs look like, those seem fine.

Merging now, thanks. --Andy

PR #47, opened 2026-04-07
DLSR-269: catch Filemaker errors in batch update script

ztucker4:
Implements [DLSR-269](https://uclalibrary.atlassian.net/browse/DLSR-269)

Updates `filemaker_batch_update.py` to continue processing if a `FileMakerError` occurs when updating a record. This is done with a try-except block in `_process_record()`. For now, I'm catching all `FileMakerError`s, not just the error 509 I saw recently - given the large batches processed by this script, I think it makes sense to deal with other errors in this way too.

I also made a small change to the value returned by `_process_record()`, which is used for reporting statistics. If an update fails, this will not count toward the total number of updated records.

Testing: 33 automated tests still pass (`docker compose exec ftva_data python -m unittest`). The code updated in this PR only executes when data is actually updated, so running with the `--dry_run` flag is not relevant.

[DLSR-269]: https://uclalibrary.atlassian.net/browse/DLSR-269?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/filemaker_batch_update.py
+++ b/filemaker_batch_update.py
@@ -426,7 +426,8 @@ def _process_record(
             # Log the error and continue processing other records
             logger.error(
                 f"Skipping record_id={record_id} inventory_id={inventory_id!r} "
-                f"due to Filemaker error: {e}"
+                f"due to Filemaker error: {e}. "
+                f"Pending changes were: {pending_changes}"
             )
             # Return 0 here to show that no changes were made
             return 0
--- a/filemaker_batch_update.py
+++ b/filemaker_batch_update.py
@@ -418,14 +418,28 @@ def _process_record(
         pending_changes[field_name] = new_value
 
     if pending_changes and not dry_run:
-        success = fm_client.edit_record(record_id=record_id, field_data=pending_changes)
+        try:
+            success = fm_client.edit_record(
+                record_id=record_id, field_data=pending_changes
+            )
+        except FileMakerError as e:
+            # Log the error and continue processing other records
+            logger.error(
+                f"Skipping record_id={record_id} inventory_id={inventory_id!r} "
+                f"due to FileMaker error: {e}"
+            )
+            # Return 0 here to show that no changes were made
+            return 0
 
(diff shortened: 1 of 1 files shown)


henry-r18 (commented):
@ztucker4 These changes look good to me. I agree that it makes sense to catch all FM exceptions not handled in `ftva-etl`, not just 509.

Maybe we should add a `records_failed` or `updates_failed` tally to the `stats` returned by `_process_batch`? I suppose it's implied by the difference between `records_processed` and `records_updated`, but could be useful to see at a glance if a batch includes some failed records that require special attention.

ztucker4:
Thanks @henry-r18 - I like the idea of reporting out the failure count directly, and I've updated the script to count and log this along with the other statistics.

filemaker_batch_update.py
```diff
@@ -418,14 +418,28 @@ def _process_record(
...
 
     if pending_changes and not dry_run:
-        success = fm_client.edit_record(record_id=record_id, field_data=pending_changes)
+        try:
+            success = fm_client.edit_record(
+                record_id=record_id, field_data=pending_changes
+            )
+        except FileMakerError as e:
+            # Log the error and continue processing other records
+            logger.error(
```
akohler, line 427: @ztucker4 I think it would be useful to log `pending_changes` here too.  The field-by-field old/new values are logged in DEBUG mode as the changes are prepared, which is fine, but if the actual update fails, knowing exactly what data was sent via API will be helpful in the error context.

akohler (changes requested):
@ztucker4 One minor change, to log the `pending_update` values if an error does occur during update, then this can be merged.
Thanks --Andy

ztucker4:
Thanks @akohler! Good catch - I've added the `pending_changes` value to the logged message.

akohler (approved):
@ztucker4 Merging now, thanks.
@henry-r18 FYI

PR #48, opened 2026-04-09
DLSR-245: Batch update `director`

henry-r18:
Implements [DLSR-245](https://uclalibrary.atlassian.net/browse/DLSR-245)

### Acceptance criteria
- [X] Extend `filemaker_batch_update.py` to apply the cleanup rules for director, as per [these specs](https://uclalibrary.atlassian.net/wiki/x/BwBRZw)
- [X] The script should allow users to specify which field is being cleaned up

### Description
This PR adds functionality for batch updating the `director` field in Filemaker. Per the specs linked above, the script should first handle delimiters in the FM values, returning the value as-is if it contains more than one type of delimiter (e.g. both comma and semi-colon). This logic is handled (maybe a little awkwardly) in `_split_multivalue_field`. After that, the transformers apply on the individual split values from FM, before being rejoined with `, `:
- `_trim_whitespace`: straightforward trimming
- `_parse_credited_names`: this handles the "5. Pseudonyms and Credited Names" section of the specs
- `_standardize_director_name`: <del>this handles initials and particles in the middle of names (e.g. `de la` and `van der`). For transparency, I used a Cursor Agent pretty liberally for help with this function and its helpers. I liked its more robust and flexible solution, especially for keeping initials capitalized, better than my initial attempts. It also came up with a nice list of `DIRECTOR_NAME_PARTICLES`.</del> **Update**: this now applies a sequence of evaluations and string substitutions mainly using regular expressions, to cover the capitalization rules, handle quoted nicknames, roman numerals, lowercasing `c/o`, and handling numbered names. It's more fully commented than it was previously, and I think it reads more clearly that the previous iteration.
- Then finally known mappings are applied to handle "4. Indeterminate and null values"

I adjusted the `timeout` param on the `FilemakerClient` to 240 from 120, because I was running into timeouts again.

I also bumped the version of `ftva-etl` to `v0.3.1` so the new FM layout is targeted.

#### Refactors to `_split_multivalue_field()`
I refactored `_split_multivalue_field()` to apply different splitting rules on a per-field basis, rather than per delimiter enum. This function is probably a good candidate for a more general refactor to make it more modular and generic, but for the time being, I created a conditional branch for `director`, which requires particular logic for splitting up most multi-value inputs, while leaving some untouched.

Rules 1, 3, and 9 from the standardization rules are handled here, rather than in the functions listed in `TRANSFORMERS`.

### Testing
I added tests to cover the various examples included in the specs, plus some others I came up with. Brings the total to 34 tests.

My `dry_run` of the script took longer than the other fields so far--nearly 2 hours: 

`python filemaker_batch_update.py -c config_dev.toml -f director --dry_run`

Here's my output summary:

`Processing complete. Processed 598686 Filemaker record(s). Would update 448208 record(s) with 448208 total field change(s).`

The large number of records updated is attributable to mapping empty strings `''` to `Unknown`.

#### Update
My latest `dry_run` took a little less than an hour. Here's the summary:

`Processing complete. Processed 598699 Filemaker record(s). Would update 445914 record(s) with 445914 total field change(s).`

The discrepancy in update counts almost certainly has to do with rule changes, as there are more values being left untouched now, though not that many relatively speaking.

I also fixed the issue with the print statement for the final iteration in the logs, so now it reads `Pagination complete. All records retrieved.` after the last page is fetched from FM.

[DLSR-245]: https://uclalibrary.atlassian.net/browse/DLSR-245?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/filemaker_batch_update.py
+++ b/filemaker_batch_update.py
@@ -17,7 +17,7 @@ from fmrest.record import Record
 logger = logging.getLogger(Path(__file__).stem)
 
 # These are delimiter variations used in Filemaker multi-value fields
-FM_DELIMITER_PATTERN = r"\s*(?:[,;/|]|\band\b|&|\r)\s*"
+FM_DELIMITER_PATTERN = r"\s*([,;/|]|\band\b|&|\r)\s*"
 
 
 # --------------------
@@ -435,7 +435,11 @@ def _split_multivalue_field(
     :param delimiter: The delimiter to split on.
     :return: A list of individual values, and a boolean indicating whether to skip transformers.
     """
-    value = value.strip()
+    # Strip leading and trailing whitespace,
+    # then also strip any leadin or trailing
+    # delimiter characters defined for the field.
+    # This prevents bad delimiters from being included in output values.
+    value = value.strip().strip(delimiter)
     if not value:
         return [""], False
 
--- a/filemaker_batch_update.py
+++ b/filemaker_batch_update.py
@@ -16,6 +16,39 @@ from fmrest.record import Record
 # handlers are configured via `_configure_logging`.
 logger = logging.getLogger(Path(__file__).stem)
 
+# These are delimiter variations used in Filemaker multi-value fields
+FM_DELIMITER_PATTERN = r"\s*(?:[,;/|]|\band\b|&|\r)\s*"
+
+DIRECTOR_NAME_PARTICLES = (
+    "de",
+    "la",
+    "le",
+    "du",
+    "des",
+    "van",
+    "von",
+    "ver",
+    "ten",
+    "ter",
+    "den",
+    "der",
+    "da",
+    "di",
+    "del",
+    "della",
+    "delle",
+    "degli",
+    "dell",
+    "dos",
+    "das",
+    "do",
+    "el",
+    "al",
+    "los",
+    "las",
+    "y",
+)
+
 
 # --------------------
 # Helper functions
@@ -140,12 +173,20 @@ MAPPINGS = {
         "N/A": "No linguistic content",
         "NONE": "No linguistic content",
     },
+    "director": {
+        "NO DIRECTOR LISTED": "N/A",
+        "N/A": "N/A",
+        "NULL": "Unknown",
+        "UNKNOWN": "Unknown",
+        "": "Unknown",
+    },
 }
 
 
 class FieldDelimiters(StrEnum):
     production_type = "\r"
     Language = ", "
+    director = ", "
 
 
 def _trim_whitespace(value: str) -> str:
@@ -181,6 +222,86 @@ def _make_uppercase(value: str) -> str:
 def _make_capitalized(value: str) -> str:
     """Convert the provided value to capitalized case."""
     return capwords(value)
+
+
+def _is_initials_token(token: str) -> bool:
+    """True if token is only single-letter initials separated by dots (e.g. A.B. or c.d.e.)."""
+    compact = token.replace(" ", "")
+    if not compact or not all(c.isalpha() or c == "." for c in compact):
+        return False
+    segments = [s for s in compact.split(".") if s]
+    return bool(segments) and all(len(s) == 1 and s.isalpha() for s in segments)
+
+
+def _format_initials(token: str) -> str:
+    """Normalize an initials token to the form A.B.C."""
+    letters = [c for c in token if c.isalpha()]
+    return ".".join(c.upper() for c in letters) + "."
+
+
+def _standardize_name_token(raw: str, word_index: int, n_words: int) -> str:
+    """Apply per-token rules inside one whitespace-delimited word (hyphens preserved)."""
+    hyphen_chunks = raw.split("-")
+    out_chunks = []
+    for chunk in hyphen_chunks:
+        if _is_initials_token(chunk):
+            out_chunks.append(_format_initials(chunk))
+        elif (
+            n_words > 1
+            and 0 < word_index < n_words - 1
+            and len(hyphen_chunks) == 1
+            and chunk.lower() in DIRECTOR_NAME_PARTICLES
+        ):
+            out_chunks.append(chunk.lower())
+        else:
+            out_chunks.append(chunk.capitalize())
+    return "-".join(out_chunks)
+
+
+def _standardize_director_name(name: str) -> str:
+    """Return a standardized format for director names, according to specs:
+
+    - Initials stay uppercase with dots between letters (e.g. A.B., C.D.E.).
+    - Common particles between first and last name stay lowercase (e.g. de la, van der).
+    - All other word parts use simple title case.
+
+    :param name: Raw name string.
+    :return: Standardized name string.
+    """
+    words = name.split()
+    n = len(words)
+    return " ".join(_standardize_name_token(w, i, n) for i, w in enumerate(words))
+
+
+def _parse_credited_names(value: str) -> str:
+    """Parse credited names from the provided value, according to specs:
+
+    - Use name following "as", e.g. Lew Landers (as Louis Friedlander) -> Louis Friedlander
+    - Use name before "i.e.", e.g. William Goodrich [i.e. Roscoe Arbuckle] -> William Goodrich
+    - Use name before "aka", e.g. Marcus aka Sid Marcus -> Marcus
+    - Note that "i.e." is normalized in `_split_multivalue_field` to avoid splitting on the comma
+
+    :param value: Raw value string.
+    :return: Parsed credited names string.
+    """
+    # Remove square brackets and parentheses and lowercase the value
+    value = re.sub(r"[\[\]\(\)]", "", value.lower())
+
+    if " as " in value:
+        parts = value.split(" as ")
+        if len(parts) == 2:
+            return parts[1].strip()
+
+    if "i.e." in value:
+        parts = value.split(" i.e. ")
+        if len(parts) == 2:
+            return parts[0].strip()
+
+    if " aka " in value:
+        parts = value.split(" aka ")
+        if len(parts) == 2:
+            return parts[0].strip()
+    return value
 
 
 def _remove_intertitles(value: str) -> str:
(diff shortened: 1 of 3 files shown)


akohler:
Just noting a few things outside of the code review: I did the same dry run, almost exactly the same numbers (448207 records & fields instead of 448208), probably due to manual editing.  However, my run took a much more reasonable time of 51 minutes, which is in line with dry runs for production type and language.

The vast majority of `Unknown` updates look correct, but there are a few noticeable insertions of `Unknown` which I think are incorrect, and could probably be avoided.  These seem to be caused by delimiters being used when there aren't multiple values to delimit, like:
```
from=', Wally Bulloch' to='Unknown, Wally Bulloch'
from='Lee R. Bobker;' to='Lee R. Bobker, Unknown'
```

Here's a full list of all `Unknown` updates:
```
# Change log filename to your own
# Grab the UPDATE log messages, look for the Unknown updates;
# split lines on space, taking the 8th field to the end (the from/to parts);
# sort them alphabetically, report counts for each unique line,
# then finally sort those numerically by count descending.

grep UPDATE logs/filemaker_batch_update_20260409_174836_DRY_RUN.log \
| grep Unknown | cut -d" " -f8- | sort | uniq -c | sort -nr

 414705 from='' to='Unknown'
   1336 from='\r' to='Unknown'
    749 from=' ' to='Unknown'
    410 from='\r\r' to='Unknown'
    326 from='unknown' to='Unknown'
     73 from='\xa0' to='Unknown'
     58 from='  \xa0' to='Unknown'
     55 from='\r\r\r\r\r\r\r' to='Unknown'
     36 from='\r\r\r' to='Unknown'
     27 from='\r\r\r\r' to='Unknown'
     16 from='\t' to='Unknown'
     14 from=' \r' to='Unknown'
     13 from=' \xa0' to='Unknown'
      8 from='\r\r\r\r\r' to='Unknown'
      2 from=' \r\r \r\r\r\r' to='Unknown'
      2 from='\r\r\r\r\r\r\r\r' to='Unknown'
      2 from='\r\r\r\r\r\r\r\r\r\r\r\r\r\r\r\r\r\r\r' to='Unknown'
      1 from=', Wally Bulloch' to='Unknown, Wally Bulloch'
      1 from='Unknown; John Lawson; Con Pederson' to='Unknown, John Lawson, Con Pederson'
      1 from='Unknown/Henry Bronchtein\r' to='Unknown, Henry Bronchtein'
      1 from='Tom Cherones/unknown/Mel Damski' to='Tom Cherones, Unknown, Mel Damski'
      1 from='\r \r\r\r\r' to='Unknown'
      1 from='\r\r\r\r\r\r' to='Unknown'
      1 from='\r\r \r\r\r\r' to='Unknown'
      1 from='Richard Gerdau ;' to='Richard Gerdau, Unknown'
      1 from='NICHOLAS CHESLA, CINDY CREAGER, AND JULIE ENGLANDER' to='Nicholas Chesla, Cindy Creager, Unknown, Julie Englander'
      1 from='Nedeljko Dragić,' to='Nedeljko Dragić, Unknown'
      1 from='MONICA NOLAN, JULIE DOFT, AND JENNI OLSON' to='Monica Nolan, Julie Doft, Unknown, Jenni Olson'
      1 from='Lee R. Bobker;' to='Lee R. Bobker, Unknown'
      1 from='Gil Martin, ' to='Gil Martin, Unknown'
      1 from='DR. LAU, ANGELA PHONG, AND MISS BEA' to='Dr. Lau, Angela Phong, Unknown, Miss Bea'
      1 from='Dave Fleischer,' to='Dave Fleischer, Unknown'
      1 from='Clyde Geronimi,' to='Clyde Geronimi, Unknown'
      1 from='Ante Zaninović;' to='Ante Zaninović, Unknown'
```

There are nearly 14,000 unique changes being made:
```
$ grep UPDATE logs/filemaker_batch_update_20260409_174836_DRY_RUN.log | cut -d" " -f8- | sort | uniq -c | wc -l

13801
```

filemaker_batch_update.py
```diff
@@ -183,6 +224,94 @@ def _make_capitalized(value: str) -> str:
...
+
+
+def _format_initials(token: str) -> str:
+    """Normalize an initials token to the form A.B.C."""
+    letters = [c for c in token if c.isalpha()]
+    return ".".join(c.upper() for c in letters) + "."
+
+
+def _standardize_name_token(raw: str, word_index: int, n_words: int) -> str:
+    """Apply per-token rules inside one whitespace-delimited word (hyphens preserved)."""
```
akohler: This needs significantly clearer documentation.  What are "per-token rules"?  What's a `token` in this context?  The method name says `name_token` but nothing like that is passed into the method.  What are `word_index` and `n_words` parameters?  How is `name` relevant as opposed to some other kinds of values developers might want to pass to this method?

The individual comments in the method help somewhat... but then everything gets muddier by some of the terminology.   What are `particles`?  They seem to be substrings, I guess? But the input string was split into `chunks`, that's what's being iterated over, Is a `particle` a `chunk`?

I don't mind AI being used as a tool - old Grumpy Gramps has slogged through lots of Stack Overflow articles and blog posts to get help with code - but the code then has to be adapted to the context of the problem and the project.  If it was confusing to write, it often will be confusing to read, especially by maintenance programmers who don't have the initial context of the prompt used to generate this.

filemaker_batch_update.py
```diff
@@ -270,19 +399,64 @@ def _normalize_language_spelling(value: str) -> str:
...
+    if new_value != value:
+        logger.debug(f"Normalized delimiters: {value!r} -> {new_value!r}")
+    return new_value
+
+
+def _split_multivalue_field(
+    field_name: str, value: str, delimiter: str
+) -> tuple[list[str], bool]:
+    """Splits a multi-value field into a list of values, based on the provided delimiter.
+    Also trims whitespace from each value and filters out any empty values.
```
akohler, line 428: There seems to be a bug in this function, as it does not match what this comment says.  I called it with one of the values noted in my big general comment, a name followed by space and then a delimiter: `"Richard Gerdau ;"`

```
from filemaker_batch_update import _split_multivalue_field
_split_multivalue_field("director", "Richard Gerdau ;", ";")
(['Richard Gerdau', ''], False)
```
From the comment (and my general assumptions), I expected a list with one value, just the name.  There's a second value, an empty string.

This does seem to be a general bug, not director-specific - the same thing happens with language:
```
_split_multivalue_field("Language", "Richard Gerdau ;", ";")
(['Richard Gerdau', ''], False)
```
If this is correct behavior and other transformers should handle the empty value, the comment should be changed.  Otherwise, the bug should be fixed to match the described behavior.

tests/test_filemaker_batch_update.py
```diff
@@ -63,6 +63,75 @@ def setUp(self):
             ("?", "Undetermined"),  # special case, known value
             ("Russian Intertitles", "Russian"),  # "Intertitles" should be removed
         ]
+        self.test_director_values = [
```
akohler, line 71: Test values should be added to confirm that the extra `Unknown` problem is fixed - see my general comment about those.  That is of course assuming those extra values are not wanted. I understand converting a standalone value which is just whitespace to `Unknown`, but I suspect that's not intended for poorly-delimited real values like `"Richard Gerdau ;"`.

akohler (changes requested):
@henry-r18 Some changes need to be made to simplify, or better comment, the generated code.  It needs to fit into the overall context of the application. Mismatches between variable names and comments must be cleaned up too.

The source data is really messy - no surprises - so I focused on what I noticed and listed in my first non-review comment, the `Unknown` values.  There are very few apparent errors with those, and I am assuming they are errors based on my reading of the specs and just what seems right, so if any doubt, check with Thelma (a rule to live by).  If these are errors, tests need to be added to ensure the fix is correct.

I did not review the sequence of transformations, as the code overflowed my mental stack.  Once that's improved, I'll review again with the full context in mind.

Thanks --Andy

henry-r18:
@akohler Thanks for your comments on this. With fresh eyes, I can see how unclear a lot of the code is, especially the generated parts. Will work on simplifying where possible, as well as clarifying variable names, comments, and doc-strings.

After [discusssion](https://uclalibrary.slack.com/archives/C0893DA0TL0/p1775764255104829) with Thelma on Slack, it sounds like the standardization rules for this are in flux a bit. I'll keep working with Thelma next week to iron things out, but it will probably involve reversing course on some of the current code, as well as adding new functions.

henry-r18:
@akohler I pushed up my latest changes, which cover the revised standardization rules and (hopefully) make the code clearer and better documented. I updated the original comment on the PR with some summary info about the changes.

Thelma signed off on the logs from my latest `dry_run`, so I'll go ahead and run against FM after your review, provided there are no more logical changes to be made. Thanks!

**Edit**: @ztucker4 could you take a look at this, when you have a chance? (Sorry to bug you on your day-off, Andy! Slipped my mind. Enjoy!)

akohler (changes requested):
@henry-r18 This revised code is much clearer, thank you.  There's still a lot going on, but it's a complex set of operations on messy data. Thelma's OK is good enough for me :)

I've added @ztucker4 as some of the changes touch language-related code, so I'd like her to sign off on those at least.

I am requesting changes though, because I found a possible problem when playing around with the test data.  There are plenty of good tests, and everything present does pass, but most of the test data checks just one problem.  What if there are multiple problems?

For example, I added this:
```
            (
                "John Smith, Tom Jones,",
                "John Smith, Tom Jones",
            ),  # two basic names with a trailing delimiter
```
and got this:
```
AssertionError: 'John Smith, Tom Jones,' != 'John Smith, Tom Jones'
- John Smith, Tom Jones,
?                      -
+ John Smith, Tom Jones
```
I don't know if this is an order of operations problem or something else.  There are passing tests for single names with a trailing (or leading) delimiter, but this fails.  Since this is not an unlikely scenario, please investigate and fix.

Then please add a few more test cases with likely combinations, focusing on ones which test the order of transformations.  These don't need to cover every possible permutation, but see if there are similar combinations which might also require tweaking.

Thanks --Andy

henry-r18:
@ztucker4 I pushed up a couple small changes to address the issue with trailing commas on the `director` field (which I think might affect `Language` too). Since the target delimiter for the `director` field (i.e. the delimiter we want to use in the output) is a comma plus a space (`, `), and the code splits multi-valued inputs on this delimiter, it was not splitting values with trailing commas, because the white-space was getting stripped prior to the split. My solution is to also strip the delimiter at the start of `_split_multivalue_field`, in addition to stripping white-space.

I also made a small tweak to `FM_DELIMITER_PATTERN`, because an input test value with two consecutive delimiters was getting mistaken as two different delimiters, and so the `director` branch was returning the value as-is. For example, `"John Smith,, Tom Jones"` (with two commas in the middle) was getting matched as one comma (`,`) followed by one comma-plus-space (`, `), and the code was treating these as two distinct delimiters.

Then I pushed a change that moves the initial capitalization logic to fix an issue I identified by adding a new test to cover the order of transformations. Anyway...tests are all passing at this point. Should be ready for review, when you have a chance.

ztucker4 (approved):
This all looks good to me. Tests and partial dry runs show expected results. I like the way `_split_multivalue_field()` and `_normalize_multivalue_delimiters()` are implemented now, and the structure of the script will work well with the Dates work I'm doing now.

henry-r18:
Thanks @ztucker4! I'm glad the changes don't interfere with your work on the Dates field.

@akohler if everything looks good to you, I'll plan to run this in production at the end of the day today.

akohler (approved):
@henry-r18 This latest version is excellent.  I ran two full dry runs, one with this code and one with code as of aa93bf7ab6c2d7299544ac88112f4f7098128c4b (the code I last reviewed).  I then compared the UPDATE statements. spot-checking.  As you noted, this fixes the double-delimiter issue which I had not seen. In particular a lot of records with ` // ` now are updated correctly.

Everything I looked at is clean.  Merging - go forth and update.

Thanks also to @ztucker4 for your review, I appreciate it.

PR #49, opened 2026-04-09
DLSR-277: Add shim to handle inventory number inconsistencies

henry-r18:
Implements [DLSR-277](https://uclalibrary.atlassian.net/browse/DLSR-277)

### Acceptance criteria
- [X] `generate_metadata.py` is updated to handle known inventory number inconsistencies between DD and Alma

### Description
Adds a shim to retry the Alma search with three known suffixes `T`, `M`, and `R` for DD inventory numbers that don't match any Alma call numbers. For example, if `VA15789` does not result in a matched Alma record, it will be retried as `VA15789T` and so on.

### Testing
This shim allowed [Batch 1](https://docs.google.com/spreadsheets/d/1LAbSnTm7YCvkJRrerJKReAvWKyrGKQYNkH-ZecN4FvM/edit?usp=sharing) to be regenerated with all records matching Alma items.

[DLSR-277]: https://uclalibrary.atlassian.net/browse/DLSR-277?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/generate_metadata.py
+++ b/generate_metadata.py
@@ -177,10 +177,33 @@ def _process_input_data(
             if bib_records:
                 bib_record = bib_records[0]
             else:
-                logger.info(
-                    f"No Alma bib records found for inventory number '{inventory_number}' "
-                    f"on record {row['dl_record_id']}. Proceeding with DD and FM data only."
-                )
+                # If no Alma bib record is found, retry with suffixes "T", "M", and "R".
+                # NOTE: This is a shim to deal with inconsistencies with inv numbers across systems.
+                for suffix in ["T", "M", "R"]:
+                    logger.info(
+                        f"No Alma bib records found for inventory number '{inventory_number}' "
+                        f"on record {row['dl_record_id']}. Retrying with suffix '{suffix}'."
+                    )
+                    all_bib_records = alma_sru_client.search_by_call_number(
+                        inventory_number + suffix
+                    )
+                    bib_records = filter_by_inventory_number_and_library(
+                        all_bib_records, inventory_number + suffix
+                    )
+                    if bib_records:
+                        logger.info(
+                            "Found Alma bib record for "
+                            f"inventory number '{inventory_number + suffix}' "
+                            f"(note suffix '{suffix}') on record {row['dl_record_id']}. "
+                            "Proceeding with DD, FM, and Alma data."
+                        )
+                        bib_record = bib_records[0]
+                        break
+                if not bib_record:
+                    logger.warning(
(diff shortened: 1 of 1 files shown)


akohler:
@henry-r18 Almost all of the code in this PR is for director updates in `filemaker_batch_update.py` and is not relevant to the inventory number shim.

Please update this PR to apply only the relevant inventory number change, or open a new PR targeting that change.
Thanks --Andy

henry-r18:
@akohler Whoops...I forked off the wrong branch. Just pushed a fix which should now target `main`, with only the inventory number shim.

akohler (changes requested):
@henry-r18 Since this adds considerable complexity & 20+ lines of code just for Alma, it's best moved into a separate helper method in the `generate_metadata` script.  That way, there will be just a few lines each for getting the relevant record for each of the 3 data sources.

That method can also eliminate the need for duplicative code to do the search and the filtering.  One option is something like this, which includes everything I needed in the REPL, and leaves out logging etc.  There may be a cleaner way to do it; my approach was just to treat the plain inventory number the same as "it has a suffix too, just the empty string".
```
from ftva_etl import AlmaSRUClient
alma_sru_client = AlmaSRUClient()
suffixes = ["", "T", "M", "R"]
inventory_number = "VA15789"
inv_nos_to_check = [inventory_number + suffix for suffix in suffixes]

bib_record = None
for inv_no in inv_nos_to_check:
    bib_records = alma_sru_client.search_by_call_number(inv_no)
    if bib_records:
        # QAD demo only; filtering step skipped here
        print(f"Found {len(bib_records)} for {inv_no}")
        bib_record = bib_records[0]
        break
if not bib_record:
    print(f"Nothing found for {inventory_number} with {suffixes}")
```
I think that with comments, it would be clear enough, while moving all the calls to search & filter to be within the loop instead of both outside and inside.  Feel free to modify or look for cleaner ways to achieve this.

Thanks --Andy

henry-r18:
@akohler Thanks for the snippet; I agree the suffix handling code is clearer in its own function. What do you think of this refactor?

akohler (approved):
@henry-r18 This refactor looks fine to me, and I got what looks like the correct record testing it with `VA15789`.  The code is a little longer than before, but I think it's clearer, especially within the busy `_process_input_data()` method.

Merged, thanks --Andy
@ztucker4 FYI for rebasing if needed.

PR #50, opened 2026-04-14
DLSR-244: fix bug with 'N/A' language value

ztucker4:
Fixes [DLSR-244](https://uclalibrary.atlassian.net/browse/DLSR-244)

A small change and corresponding test to correctly map the Language value “N/A” to “No linguistic content”.

Currently, the first processing step of the script is to attempt to split the multivalue Language field into individual components. This includes splitting on slashes (for cases like "English/French"), which caused the value "N/A" to split into "N, A". 

To address this, I added a case in `_split_multivalue_field()` to return the value "N/A" as is, which allows the script to correctly map it in a later step.  This seemed like the most straightforward approach, but it would not work correctly for a Language value like "English, N/A". For now, I can confirm that there are/were no instances of N/A combined with another value in the FM database.

I added one test, for a total of 33: `docker compose exec ftva_data python -m unittest`

[DLSR-244]: https://uclalibrary.atlassian.net/browse/DLSR-244?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/filemaker_batch_update.py
+++ b/filemaker_batch_update.py
@@ -277,9 +277,14 @@ def _split_multivalue_field(value: str, delimiter: str) -> list[str]:
     """Split a multi-value field into a list of values, based on the provided delimiter.
     Also trims whitespace from each value and filters out any empty values."""
     if delimiter == FieldDelimiters["Language"].value:
-        # First, check for the known value "N/A", which should not be split
-        if value.strip() == "N/A":
-            return [value.strip()]
+        # First, check for the known value "N/A.
+        # If present, replace with "NONE" and process as usual,
+        # which will eventually be mapped to "No linguistic content" by the mapping function.
+        if "N/A" in value:
+            logger.debug(
+                "Found known value 'N/A' in language field; replacing with 'NONE'."
+            )
+            value = value.replace("N/A", "NONE")
         # Normalize all possible delimiters to comma for language
         value = re.sub(r"\s*(?:[,;/|]|\band\b|&|\r)\s*", ", ", value)
         logger.debug(f"Normalized delimiters to comma: {value!r}")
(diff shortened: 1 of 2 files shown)


filemaker_batch_update.py
```diff
@@ -277,6 +277,9 @@ def _split_multivalue_field(value: str, delimiter: str) -> list[str]:
     """Split a multi-value field into a list of values, based on the provided delimiter.
     Also trims whitespace from each value and filters out any empty values."""
     if delimiter == FieldDelimiters["Language"].value:
+        # First, check for the known value "N/A", which should not be split
+        if value.strip() == "N/A":
```
akohler: A more robust way to do this: replace "N/A" with "NONE", which per the language mappings will also ultimately be mapped to "No linguistic content".  "NONE" avoids the delimiter issues.

akohler (changes requested):
@ztucker4 I suggested a more robust way to achieve the same result, though it's possible only because for Language, "NONE" and "N/A" map to the same final value.

As you noted, there currently are no values in the real data like "English, N/A",  but I think if people can enter bad data, they will.  Optional bonus: read "Been a Long, Long Time" by R.A. Lafferty, a writer of some of my favorite humorous short science fiction: https://gwern.net/doc/fiction/humor/1970-lafferty.pdf

A clunkier but more general approach: replace "N/A" with "PLACEHOLDER", split the value as usual, then replace "PLACEHOLDER" with "N/A".  That has the advantage of being extensible - any in a list of troublesome terms could be swapped for "PLACEHOLDER_NA", "PLACEHOLDER_FOOBAR", "PLACEHOLDER_HELP_ME", etc. - but I think not needed here.

ztucker4:
Hi @akohler, this is ready for another look. The code and tests are updated to map N/A to NONE before splitting, supporting values like "N/A; English" and hopefully fortifying us against a few more monkey-years :)

akohler (approved):
@ztucker4 Looks good, merging.

PR #52, opened 2026-04-22
DLSR-246: FM batch updates for dates

ztucker4:
Implements [DLSR-246](https://uclalibrary.atlassian.net/browse/DLSR-246)
Relevant specs: https://uclalibrary.atlassian.net/wiki/x/GQBOZw

Adds support for normalizing two new FM fields, `release_broadcast_year` and `record_date`, using the same logical rules. I am running a full dry run of the script now and will provide the output log file with all intended updates (and non-updates of non-empty values) in Slack tomorrow.

For a full dry run: `docker compose exec ftva_data python filemaker_batch_update.py -c prod_config_secrets.toml -f release_broadcast_year record_date --dry_run`

For tests: `docker compose exec ftva_data python -m unittest`
 
When running the script or tests, you may see a message about `tzlocal` awkwardly placed in the terminal output. I believe this is coming from one of the two date parsing libraries, but I didn't find a quick way to silence it.

This is a complicated set of requirements, and the code relies on many complex regexes (created with assistance from LLMs). To try to compensate for this opacity, I've added a large number of tests, most of which use values found in the database. I've also added the `edtf` and `dateparser` libraries, which are helpful in a limited set of cases.

[DLSR-246]: https://uclalibrary.atlassian.net/browse/DLSR-246?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/filemaker_batch_update.py
+++ b/filemaker_batch_update.py
@@ -170,8 +170,8 @@ class FieldDelimiters(StrEnum):
     production_type = "\r"
     Language = ", "
     director = ", "
-    release_broadcast_date = "\r"
-    record_date = "\r"
+    release_broadcast_year = "; "
+    record_date = "; "
 
 
 def _trim_whitespace(value: str) -> str:
@@ -410,16 +410,16 @@ def _normalize_copyright_and_circa(value: str) -> str:
     copyright_pattern = re.compile(r"c(?:opyright)?\.?\s*(\d{4})", re.IGNORECASE)
     value = copyright_pattern.sub(r"c\1", value)
 
-    # Circa: Replace circa, ca., CA., CIRCA, c. with year?
-    # ca. 1919 -> 1919?; CIRCA 1970 -> 1970?
-    # c. 1910 -> 1910?
+    # Circa: Replace circa, ca., CA., CIRCA, c. with question mark after year
+    # Exception: Convert circa date range to “or” format first
+    # circa 1929-1930 -> 1929 or 1930
+    value = re.sub(r"(?i)circa\s+(\d{4})\s*[-–]\s*(\d{4})", r"\1 or \2", value)
+
+    # Circa: Replace single-year circa variations with question mark after year
+    # e.g. circa 1970 -> 1970?
     value = re.sub(
         r"(?i)\b(?:circa|ca\.?|c\.)\s*(\d{4})\b", lambda m: f"{m.group(1)}?", value
     )
-
-    # Exception: Convert circa date range to “or” format
-    # circa 1929-1930 -> 1929 or 1930
-    value = re.sub(r"(?i)circa\s+(\d{4})\s*[-–]\s*(\d{4})", r"\1 or \2", value)
     return value.strip()
 
 
@@ -627,7 +627,7 @@ TRANSFORMERS = {
         _handle_U_X_placeholders,
         _normalize_date_ranges,
     ],
-    "release_broadcast_date": [
+    "release_broadcast_year": [
         _trim_whitespace,
         lambda value: MAPPINGS["date"].get(value, value),
         _remove_brackets,
@@ -765,6 +765,9 @@ def _initialize_client(config: dict) -> FilemakerClient:
             timeout=240,
         )
         logger.info("Connected to Filemaker.")
+        print(
+            client._fms
+        )  # Log the internal FilemakerServer instance for debugging connection issues
         return client
     except FileMakerError as e:
         logger.error(f"Failed to connect to Filemaker: {e}")
@@ -840,7 +843,7 @@ def _process_record(
         if current_value == new_value:  # skip if no change
             continue
 
-        logger.debug(
+        logger.info(
             f"UPDATE field_name={field_name} "
             f"record_id={record_id} inventory_id={inventory_id} "
             f"from={current_value!r} to={(new_value or '')!r}"  # use `!r` so `\r` is visible in log
@@ -931,6 +934,7 @@ def main() -> None:
     args = _get_arguments()
     _configure_logging(dry_run=args.dry_run)
     config = _get_config(args.config_file)
+    logger.info(config)
 
     if args.dry_run:
         logger.info("DRY RUN: no changes will be written to Filemaker.")
--- a/filemaker_batch_update.py
+++ b/filemaker_batch_update.py
@@ -413,18 +413,43 @@ def _normalize_copyright_and_circa(value: str) -> str:
     """Normalize various representations of "Copyright" and "Circa" in the provided value."""
     # Copyright: Standardize to lowercase c followed immediately by year.
     # c 1978, C1988, COPYRIGHT 2007 -> c1978, c1988, c2007
-    value = re.sub(r"(?i)c(?:opyright)?\.?\s*(\d{4})", r"c\1", value)
+    # Normalize curly apostrophes to simple ASCII
+    value = value.replace("’", "'").replace("‘", "'")
+
+    # Normalize decade forms with apostrophes (e.g. "1920's", "1920’s") to a consistent
+    # representation so subsequent rules can detect decades reliably.
+    value = re.sub(
+        r"(\d{3})0['’]s\??", lambda m: f"{m.group(1)}-?", value, flags=re.IGNORECASE
+    )
+
+    # If a circa/token prefix remains (e.g. "ca. 192-?"), drop the prefix so we
+    # end up with the normalized decade token ("192-?"). Only remove the prefix
+    # when what follows looks like a normalized decade (three digits + '-?').
+    value = re.sub(r"(?i)^(?:circa|ca\.?|c\.? )\s*(?=\d{3}-\?)", "", value)
+
+    # COPYRIGHT: Standardize to lowercase c followed immediately by year.
+    # Require a word-boundary after the year so we don't accidentally match decade tokens
+    # such as "1970s" (which should be handled above).
+    value = re.sub(r"(?i)c(?:opyright)?\.?\s*(\d{4})\b\.?", r"c\1", value)
 
     # Circa: Replace circa, ca., CA., CIRCA, c. with question mark after year
     # Exception: Convert circa date range to “or” format first
     # circa 1929-1930 -> 1929 or 1930
     value = re.sub(r"(?i)circa\s+(\d{4})\s*[-–]\s*(\d{4})", r"\1 or \2", value)
 
+    # Handle circa + decade (e.g. "c. 1970s", "ca. 1920's") -> normalize to decade uncertainty
+    value = re.sub(
+        r"(?i)(?:circa|ca\.?|c\.)\s*(\d{3})0s\b\??", lambda m: f"{m.group(1)}-?", value
+    )
+
     # Circa: Replace single-year circa variations with question mark after year
     # e.g. circa 1970 -> 1970?
     value = re.sub(
         r"(?i)\b(?:circa|ca\.?|c\.)\s*(\d{4})\b", lambda m: f"{m.group(1)}?", value
     )
+
+    # If 'circa' follows the year (e.g. '1924 circa') normalize to '1924?'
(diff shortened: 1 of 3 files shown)


ztucker4:
...and I'm already seeing a new bug in the logs (certain comma separated lists of dates, like `04-02-1959, 04-09-1959`, are being transformed to singular YYYY-MM values). I'll address this tomorrow.

filemaker_batch_update.py
```diff
@@ -384,6 +400,203 @@ def _normalize_language_spelling(value: str) -> str:
...
+    v = value.replace("[", "").replace("]", "")
+    v = v.replace("(", "").replace(")", "")
+    return v.strip()
+
+
+def _normalize_copyright_and_circa(value: str) -> str:
+    """Normalize various representations of "Copyright" and "Circa" in the provided value."""
+    # Copyright: Standardize to lowercase c followed immediately by year.
+    # c 1978, C1988, COPYRIGHT 2007 -> c1978, c1988, c2007
+    copyright_pattern = re.compile(r"c(?:opyright)?\.?\s*(\d{4})", re.IGNORECASE)
```
akohler: This is inconsistent with the other two replacement regex in this method.  The `compile` made me think `copyright_pattern` would be used repeatedly, but it isn't.  This pattern and replacement also are simpler / shorter than the other two.

I suggest replacing this 2-line compile/sub with the one-line `sub` approach used in the other 2.

filemaker_batch_update.py
```diff
@@ -384,6 +400,203 @@ def _normalize_language_spelling(value: str) -> str:
...
+def _remove_brackets(value: str) -> str:
+    """Remove square brackets and parentheses from the provided value, if present."""
+    if not value:
+        return ""
+    v = value.replace("[", "").replace("]", "")
+    v = v.replace("(", "").replace(")", "")
+    return v.strip()
+
+
+def _normalize_copyright_and_circa(value: str) -> str:
```
akohler, line 415: All 3 replacements in this method require a 4-digit year.  That's implied in the specs, but I didn't see it explicitly stated.  Does the source data support this assumption?  For example, "Copyright 202?".

No code changes may be needed here, depending on the data & Thelma's review of the dry run, but you may need to ask her explicitly about this.  Then if appropriate, add a comment to the method confirming that only 4-digit years are supported, and possibly a test case.

The "circa" replacements are less likely to be affected by this, as (sane) people are less likely to enter "circa 202?".

filemaker_batch_update.py
```diff
@@ -384,6 +400,203 @@ def _normalize_language_spelling(value: str) -> str:
...
+        r"(?i)\b(?:circa|ca\.?|c\.)\s*(\d{4})\b", lambda m: f"{m.group(1)}?", value
+    )
+    return value.strip()
+
+
+def _normalize_date(value: str) -> str:
+    """Normalize dates using EDTF (for uncertainty/unknown digits/ranges) with a
+    fallback to dateparser for natural-language dates.
+    """
+    v = (value or "").strip()
```
akohler: I'm OK with 1- or 2- letter variable names in really short blocks or methods, where it's easy to see the transient lifespan of the variable.  Here though, the method is really long, and the variable `v` tends to disappear into into the regexes and other function calls.  These also vary, so sometimes `v` is first or last or in the middle.

So, feel free to ~hate~ dislike me, but please use `value` instead of `v` throughout this method.

This will also make `_normalize_date()` consistent with the other methods where `value` is used throughout (except for `_remove_brackets()` which is OK as it's extremely short and simple).

filemaker_batch_update.py
```diff
@@ -410,6 +623,26 @@ def _normalize_language_spelling(value: str) -> str:
         _standardize_director_name,
         lambda value: MAPPINGS["director"].get(value.upper(), value),
     ],
+    "record_date": [
```
akohler, line 720: Please add a comment here to make explicit that `record_date` and `release_broadcast_year` have identical transformation sequences.  It's a limitation of using this dictionary, I don't really see anything more suitable given the field-driven approach, and that's OK.  An explicit comment here confirms for the reader / maintainer that yes, at present these really are identical by design.

filemaker_batch_update.py
```diff
@@ -610,6 +843,13 @@ def _process_record(
         new_value = _apply_transformers(field_name, current_value)
 
         if current_value == new_value:  # skip if no change
+            # If the values are not blank, log a message to aid in review
+            if current_value != "":
+                logger.debug(
+                    f"NO CHANGE field_name={field_name} "
+                    f"record_id={record_id} inventory_id={inventory_id} "
+                    f"from={current_value!r} to={(new_value or '')!r}"
```
akohler: I think the `(new_value or '')` is not needed, this can just be `new_value`.  This seems like a defense against `None`, but `_apply_transformers()` does not return `None`.

tests/test_filemaker_batch_update.py
```diff
@@ -185,6 +185,59 @@ def setUp(self):
...
+            ("1959 - 1963", "1959-1963"),
+            ("1975-76", "1975-1976"),
+            ("2015-03-12", "2015-03-12"),  # valid date, no transform needed
+            ("1939-11-28", "1939-11-28"),
+            ("JUNE 28-29", "JUNE 28-29"),  # date range of days should not be modified
+            ("MAY 9-10, 1980", "MAY 9-10, 1980"),
+            ("APRIL 30 & MAY 1-3, 1980", "APRIL 30 & MAY 1-3, 1980"),
+            ("MAY 7,8 & 9", "MAY 7,8 & 9"),
+            ("FALL 1978", "FALL 1978"),  # season and year should not be modified
+        ]
```
akohler, line 296: It seems silly, but I think there should also be test data for a plain old 4-digit year, to confirm it is not modified.  Your extensive set of test data does an excellent job of representing the messiness which should be transformed, along with complex examples which should not be updated.  Let's make sure the super-basic cases are also covered, so future changes to the code don't accidentally break those.

akohler (changes requested):
@ztucker4 This is really excellent work.  You've done a great job of documenting complex code through comments, explaining why multiple date parsing libraries are used in different scenarios, and covering all of this messiness with test cases.

I do have a few requested changes to make some of the code more consistent and maintainable.  There's a potential edge case in the copyright method, though that may just need an extra comment.  Finally, to be extra conservative, test cases should ensure the simplest already-correct years are not modified through future changes to this complex code.

As for the timezone deprecation warning: I agree this can be ignored.  I think there's no easy way to turn this one warning off, and since none of this code touches times we should be fine.  When we have exact dates, those should never change (and test cases confirm that 👍 ).  Imprecise dates won't (I think...) be changed in a TZ-related way, like rounding the date up or down depending on midnight in different time zones.

I did look into the warning, which comes from [this check](https://github.com/regebro/tzlocal/blob/master/tzlocal/unix.py#L137) in the `tzlocal` package.  Comments in that code explain a bit more about this.  The way we [set time zone in our Docker container](https://github.com/UCLALibrary/ftva-mams-data/blob/main/Dockerfile#L8) is a bit clunky, and does not update `/etc/timezone`; doing it the "correct" way in Debian (the base distro we use) seems to require `systemd` now, which is more complex than I want and might not address this anyhow.

I tested a bit in our container:
```
# In our container, /etc/timezone has the default UTC TZ
ftva_data@865b7efec821:~/project$ cat /etc/timezone
Etc/UTC

ftva_data@865b7efec821:~/project$ python
Python 3.13.5 (main, Jul 22 2025, 01:46:07) [GCC 12.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
# Run the method which emits the warning to see what it returns
>>> from tzlocal.unix import _get_localzone_name
>>> _get_localzone_name()
/etc/timezone is deprecated in some distros, and no longer reliable. tzlocal is ignoring it, and you can likely delete it.
'America/Los_Angeles'

# This value comes from `/etc/timezone`, which is set in our Dockerfile.
>>>

# Double check via shell that the TZ (for me) comes from inside the container, not my Alaska time zone
ftva_data@865b7efec821:~/project$ date
Tue Apr 21 17:45:42 PDT 2026
```

Thanks --Andy

akohler:
> ...and I'm already seeing a new bug in the logs (certain comma separated lists of dates, like `04-02-1959, 04-09-1959`, are being transformed to singular YYYY-MM values). I'll address this tomorrow.

I almost asked about lists of values in my review, then figured "nah, nobody would ever do **that**"......

filemaker_batch_update.py
ztucker4: I did find a similar example in the data (`C. 1970s`), so I'll clarify with Thelma.

ztucker4:
Hi @akohler, this is ready for another look. In addition to your suggested improvements, I added some additional logic for the [newly-documented requirements](https://uclalibrary.atlassian.net/wiki/x/GQBOZw) for decade ranges and decades with circa prefixes. I also added handling for some more strange cases I saw in the logs (lists of dates separated by `\r`, values with curly apostrophes, and "circa" appearing at the end of values).

tests/test_filemaker_batch_update.py
```diff
@@ -235,8 +237,35 @@ def setUp(self):
...
+            (
+                "C. 1970s",
+                "197-?",
+            ),  # decade with circa should be normalized to decade format
+            (
+                "ca. 1920’s?",
+                "192-?",
+            ),  # decade with circa and question mark should be normalized to decade format
+            (
+                "1960-1970s",
```
akohler, line 266: This seems odd, probably a Thelma confirmation, but based on the preceding which equates to `1920s` = `192-?`, I'd expect `1960-1970s` to go to `1960-197-?`.  The existing range implies 1960 thru 1979, with uncertainty.

filemaker_batch_update.py
```diff
@@ -573,6 +610,10 @@ def _normalize_date_ranges(value: str) -> str:
     )
     # Spaced dash: 1959 - 1963 -> 1959-1963
     value = re.sub(r"(\d{4})\s*-\s*(\d{4})", r"\1-\2", value)
+    # Range where second part is a decade (e.g. 1960-1970s) -> normalize to 1960-1970
```
akohler, line 665: Commented on the test for this, same concern: should `1960-1970s` really be truncated to `1960-1970` (as this does), which could reduce precision by as much as 9 years?  If Thelma confirms, then this is OK, but I'd add an acknowledgement of this loss in the comments so we know this is intended.

akohler (changes requested):
@ztucker4 Thanks for making the changes I requested, and for handling all the newly-discovered edge cases.

I'm not asking for any new changes, but flagged code & test for one case which seems like a significant loss of precision.  If Thelma confirms it, just add confirmation to the comments; otherwise, adjust as needed.

I also noted another possible edge case in Slack, where `/00/` as a day of month was converted to (apparently) today's day of month, `-22`.  That should be fixed and tested as well.

Once these and Thelma's feedback are incorporated, this will be ready for what will probably be another quick review and merge.

Thanks --Andy

tests/test_filemaker_batch_update.py
ztucker4: I asked Thelma in Slack about this case (`1960-1970s`), and she confirmed the mapping to `1960-1970`. The [requirements](https://uclalibrary.atlassian.net/wiki/x/GQBOZw) are now updated to include this.

ztucker4:
Hi @akohler, this is ready for a hopefully final look. Thelma has approved a log file produced by a dry run of the current version of the script, which should address all of your and her comments from Wednesday. Thanks for all your help with this, and thanks again for taking on the update run this weekend!

akohler (approved):
@ztucker4 I think this has been bludgeoned into submission.  The code, or me, not sure which :) 

All tests pass.  Thelma's happy.  The "Y2K"-related code and related tests looks like it will work correctly if we have to run this next year.

Thank you very much for your hard work on this! Merging now.

PR #53, opened 2026-05-05
DLSR-323: Improve timeouts and restarts in batch update script

akohler:
Implements [DLSR-323](https://uclalibrary.atlassian.net/browse/DLSR-323).

This PR makes 2 small improvements to `filemaker_batch_update.py`:
* Adjust Filemaker client timeouts
  * Initial timeout can now be set optionally in each user's config file, under `[filemaker]`, with `timeout=some_value`.
  * The initial timeout in the script defaults to 120 seconds if the user does not override via config.
  * When updating records, the timeout is reduced to 5 seconds (hard-coded) for each update, then immediately reset to the longer default.
* Allow starting the program at a specific offset / position within the record set.  By default this is 1.
  * Note that this is *not* `record_id`, but the position within the overall list of (all) records being retrieved.

I found that almost all timeouts happened when trying to update a record.  The script was waiting for up to 4 minutes (240 seconds, originally), the timeout would end, and the script would resume.  Reducing the timeout when trying an update allows the script to recover & resume much more quickly almost always.

The timeout problem itself may have many contributing factors: local network / machine, Filemaker server, resource leaks (API client and/or the server itself), record locking.... The script did tend to get slower, with more timeouts, as it progressed through large numbers of updates, so I suspect resource leak(s) somewhere, but did not see anything obvious in our script or a brief look at the `fmrest` package.

No new tests for these changes, but try a dry run using an offset very close to the end.  Currently, this is what I got with an offset of `598920`, which is just below the current total of `598924`:
```
docker compose run ftva_data python filemaker_batch_update.py -c prod_config_secrets.toml -f release_broadcast_year record_date --dry_run --offset 598920

DRY RUN: no changes will be written to Filemaker.
Fields to process: ['release_broadcast_year', 'record_date']
Connected to Filemaker.
Retrieving records in pages of 5000...
Retrieved records 598920 to 598924...
Pagination complete. All records retrieved.
Processing complete. Processed 5 Filemaker record(s). Would update 0 record(s) with 0 total field change(s).
```

The log should show a starting `record_id=796051`.

@henry-r18 FYI.

[DLSR-323]: https://uclalibrary.atlassian.net/browse/DLSR-323?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/filemaker_batch_update.py
+++ b/filemaker_batch_update.py
@@ -99,6 +99,13 @@ def _get_arguments() -> argparse.Namespace:
         required=False,
         help="Page size (i.e. `limit` param) for fetching records from Filemaker. Default is 5000.",
     )
+    parser.add_argument(
+        "--offset",
+        type=int,
+        default=1,
+        required=False,
+        help="Offset (position in list of records, NOT `record_id`) to start fetching records from Filemaker. Default is 1.",
+    )
     return parser.parse_args()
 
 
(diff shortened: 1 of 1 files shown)


ztucker4 (approved):
This looks good to me. I ran the suggested `--dry-run` on the last few records in the database to confirm that `--offset` is working correctly. (I was briefly confused by the indexing starting at 1 instead of 0, but this is consistent Filemaker behavior and I don't think it needs to be called out anywhere). The timeout changes make sense given the server behavior described in the PR description. Merging, thanks!

PR #54, opened 2026-05-20
DLSR-329: FM rules validation report

ztucker4:
Implements [DLSR-329](https://uclalibrary.atlassian.net/browse/DLSR-329) 

Adds a new script, `filemaker_validation_report.py`, that checks each FM record (in a given layout and modification timeframe) against the validation rules defined by the FTVA team ([Google Docs link](https://docs.google.com/document/d/11vt7XdENaFjr-i0IXODXvwFjYK41Fpu-IHVxke9ftEg/edit?tab=t.0)). Violations are collected into a list of dicts and written to a CSV report. Shared FM connection and record-retrieval logic is extracted to `filemaker_utils.py` to avoid duplication with the existing `filemaker_batch_update.py` script.

Sample usage (see README for more details): `python filemaker_validation_report.py --config_file prod_config_secrets.toml --layout "NEW DIGITAL_API" --start_date 05/25/2026 --end_date 06/03/2026`. A log file will be created in `logs/`, and the report CSV will be created in `reports/`.

I experienced timeouts when running the script for a range of more than about 2 weeks against either the `NEW DIGITAL_API` or `InventoryForLabeling_API` layouts. The `NEW DIGITAL STORAGE_API` contains only 2 records, and neither has any violations.

Future work and other notes:
- The script does not currently check the Notes field for violations.
- `filemaker_batch_update.py` needs to be updated to use `filemaker_utils.py`.
- The specs do not actually call for a "reason" column in the report CSV, but I found it helpful in writing and debugging the script and suspect that FTVA users will find it helpful too.

[DLSR-329]: https://uclalibrary.atlassian.net/browse/DLSR-329?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/filemaker_validation_report.py
+++ b/filemaker_validation_report.py
@@ -177,7 +177,6 @@ LAYOUT_VALIDATORS: dict[str, dict[str, list[Callable]]] = {
         # file_path and file_size are in portal_pDMUnit; see PORTAL_FIELD_VALIDATORS.
         "Barcode": [_check_null],
         "Location": [_check_null],
-        "storage_format": [_check_null],
     },
 }
 
--- a/filemaker_validation_report.py
+++ b/filemaker_validation_report.py
@@ -1,0 +1,482 @@
+import argparse
+import csv
+import logging
+import re
+import sys
+from collections.abc import (
+    Callable,
+)  # For type hinting single-field validator callables
+from datetime import datetime
+from pathlib import Path
+
+from fmrest.record import Record
+
+import filemaker_utils as fm_utils
+
+# ---------------------------------------------------------------------------
+# Module-level logger
+# ---------------------------------------------------------------------------
+logger = logging.getLogger(Path(__file__).stem)
+
+# Output columns, matching the spec's required fields plus record_id for
+# debugging convenience.
+REPORT_FIELDNAMES = [
+    "record_id",
+    "inventory_id",
+    "inventory_no",
+    "user_last_modified",
+    "date_modified",
+    "field",
+    "violation",
+]
+
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+
+def _is_null(value: str) -> bool:
+    """Return True if *value* is empty or whitespace-only."""
+    return not value or not value.strip()
+
+
+def _field(record: Record, name: str) -> str:
+    """Return a field value from *record* as a stripped string, or '' if absent."""
+    try:
+        return str(record[name] or "").strip()
+    except (KeyError, AttributeError):
+        return ""
+
+
+def _make_violation(record: Record, field: str, message: str) -> dict:
+    """Build a violation dict in the spec's output format."""
+    return {
+        "record_id": record.record_id,
+        "inventory_id": _field(record, "inventory_id"),
+        "inventory_no": _field(record, "inventory_no"),
+        "user_last_modified": _field(record, "user_last_modified"),
+        "date_modified": _field(record, "date_modified"),
+        "field": field,
+        "violation": message,
+    }
+
+
+# ---------------------------------------------------------------------------
+# Single-field validators
+# Each callable takes a string value and returns a list of error messages.
+# ---------------------------------------------------------------------------
+
+
+def _check_null(value: str) -> list[str]:
+    """Field must not be null / empty."""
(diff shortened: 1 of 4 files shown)


filemaker_validation_report.py
```diff
@@ -0,0 +1,588 @@
...
+    },
+    "NEW DIGITAL STORAGE_API": {
+        "inventory_id_field": "Inventory_id_fk",
+        "inventory_no_field": "Inventory_no_fk",
+        "user_last_modified_field": "ModifiedBy",
+        "date_modified_field": "ModificationTimestamp",
+    },
+}
+
+# Portal name on the NEW DIGITAL_API layout that contains the DC-only fields.
```
akohler: Should this be "DM-only"?

filemaker_validation_report.py
```diff
@@ -0,0 +1,588 @@
...
+PORTAL_DM_ITEMS = "portal_Portal_DM_Items"
+
+# Table occurrence prefix used by FM for fields in the DM Items portal.
+DM_PREFIX = "Digital Media_Item Unit::"
+
+# Portal name on the NEW DIGITAL STORAGE_API layout containing file path/size.
+PORTAL_DM_UNIT = "portal_pDMUnit"
+
+# Table occurrence prefix used by FM for fields in the DS portal.
```
akohler: This comment should be clearer. I think DS is Digital Storage, but then the constant name refers to Carrier and the value refers to Media.

filemaker_validation_report.py
```diff
@@ -0,0 +1,588 @@
...
+                            layout,
+                            record,
+                            date_field,
+                            f"{date_field} is null but retention_status is "
+                            f"'Temporary'.",
+                        )
+                    )
+
+        # Rule 4: Preservation/access records require sound_format.
+        OBJECT_PURPOSES_REQUIRING_SOUND = {"PRESERVATION (FULL)", "ACCESS (CORE)"}
```
akohler: Why use a set here, instead of a list?

filemaker_validation_report.py
```diff
@@ -0,0 +1,588 @@
...
+        return ["Field is null or empty."]
+    return []
+
+
+# ---------------------------------------------------------------------------
+# Layout -> flat field -> validators mapping
+# ---------------------------------------------------------------------------
+# Fields that live directly on the record (not in portals).
+
+LAYOUT_VALIDATORS: dict[str, dict[str, list[Callable]]] = {
```
akohler, line 141: I'm curious why this uses a list of Callable? I don't see any using more than one, `_check_null`.  Do the "cross-field rule also applies" comments mean this was a list in initial implementation but evolved away from that?

I don't oppose using a list here, for flexibility, and `validate_record()` iterates over them. that's all fine, but the mix of single-value lists and comments referring to other rules confuses me.

akohler (changes requested):
@ztucker4 This is excellent. I really appreciate the effort that went into understanding "portals" and the flexible design of this program.  Great use of comments capturing the requirements and the thinking behind the implementation.

I noted a few places where the comments should be clearer with abbreviations, where the info might be elsewhere in the program but should be either repeated or brought into the context of the abbreviations.  For my question about the list of Callables in the validators, I think the implementation is fine, but needs more explicit documentation of "why".

After the `ftva-etl` package update, I ran the script against "NEW DIGITAL_API" for June 1-5:
> Processed 226 record(s); 808 violation(s)

This will keep FTVA folks busy.... 

Thanks --Andy

filemaker_validation_report.py
ztucker4: No idea - I think this was just a typo that happened to work the same as a list.

filemaker_validation_report.py
ztucker4: Yes, this was a holdover from an earlier plan to use `LAYOUT_VALIDATORS` for single-field, cross-field, and portal checks. I don't know if we'll ever need more than one per field, but I've added a comment about how using lists makes it easy to add more.

akohler (approved):
@ztucker4 Thanks for making the changes, they all look great - merging now.

PR #55, opened 2026-06-08
DLSR-375: use digital_record field to filter NDM data in FM validation report

ztucker4:
Implements [DLSR-375](https://uclalibrary.atlassian.net/browse/DLSR-375)

Updates `filemaker_validation_report.py`to use the `digital_record` field to distinguish between NDM carrier (`NEW DIGITAL_API`) and legacy/GE (`InventoryForLabeling_API`) data. The previous version of this script only used the `layout` parameter in the client, which still found records associated with both types of data.

Sample output for the date range 6/1/26-6/8/26 was sent to Thelma in Slack. To run it yourself: `docker compose exec ftva_data python filemaker_validation_report.py --config_file prod_config_secrets.toml --layout "NEW DIGITAL_API" --start_date 06/01/2026 --end_date 06/08/2026`

[DLSR-375]: https://uclalibrary.atlassian.net/browse/DLSR-375?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/filemaker_validation_report.py
+++ b/filemaker_validation_report.py
@@ -436,6 +436,33 @@ def _write_csv_report(violations: list[dict], output_path: Path) -> None:
 # ---------------------------------------------------------------------------
 
 
+def _build_query_criterion(
+    layout: str,
+    date_field: str,
+    start_date: str | None,
+    end_date: str | None,
+) -> dict | None:
+    """Build the FileMaker find query criterion dict for the given layout and args.
+
+    Returns a dict suitable for use as a single find criterion, or None if no
+    find query is needed.
+
+    AND conditions within a find request are expressed by putting multiple
+    field-value pairs in the same dict. The digital_record filter is always
+    included for GE and DC layouts, combined with the date range if provided.
+    """
+    digital_record_filter = LAYOUT_DIGITAL_RECORD_FILTER[layout]
+    criterion: dict = {}
+
+    if start_date and end_date:
+        criterion[date_field] = f"{start_date}...{end_date}"
+
+    if digital_record_filter is not None:
+        criterion["digital_record"] = digital_record_filter
+
+    return criterion if criterion else None
+
+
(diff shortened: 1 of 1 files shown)


filemaker_validation_report.py
```diff
@@ -435,18 +446,34 @@ def _get_records_for_layout(
...
         )
         logger.info(f"[{layout}] Found {len(records)} record(s) in date range.")
+    elif digital_record_filter is not None:
+        # No date range, but still need to filter by digital_record.
+        logger.info(
+            f"[{layout}] No date range supplied; retrieving all records "
+            f"with digital_record={digital_record_filter!r}."
+        )
+        records = fm_client.find_all_records(
+            query=[{"digital_record": digital_record_filter}],
```
akohler: I'd like this to be consistent across the two filtered branches.  Build `query_criteria`, then call `find_all_records` once with `query=[query_criteria]`.  This might be clearer by extracting some of the setup into small helper methods.  At present these are inconsistent in logging too: the date branch logs record count, the new digital record filter branch does not.

akohler (changes requested):
@ztucker4 The new filtering functionality is good.  I'd like the two filtered branches to be consistent with each other, which I think can be done by breaking this down into small but consistent steps.  There are two checks for `digital_record_filter is not None`, two calls to `find_all_records` which vary only in the value of the `query` parameter.  The pre-fetch logging message could be made consistent, "Fetching records with criteria: ..." for example.  There should be one consistent post-fetch message, probably just a "layout found n records" message right before returning, no need to repeat the criteria since there's already a logging message for those.

Thanks --Andy

akohler (approved):
@ztucker4 This is much cleaner, thank you.  Merging now.

PR #56, opened 2026-06-11
DLSR-371: refactor batch update script w/shared utils

ztucker4:
Implements [DLSR-371](https://uclalibrary.atlassian.net/browse/DLSR-371)

Moves the shared utils files (`spacy_utils.py` and `filemaker_utils.py`) into a new `utils` directory with an `__init__.py` for use as a package. Updates `filemaker_batch_update.py` to replace several previously duplicated functions with imports from `filemaker_utils.py`.

Also includes a small change to the imports in `filemaker_validation_report.py` to reflect the new package structure.

The `__init__.py` file is minimal, containing only a docstring. This is different from the approach we took in the `archivesspace-toolkit` repo (https://github.com/UCLALibrary/archivesspace-toolkit/blob/main/python/utils/__init__.py), where we re-exported utilities functions into the generic `utils` package. In the Aspace repo, this allowed for imports to be written like `from utils import configure_logging`. For this repo, imports must be written like `from utils.filemaker_utils import configure_logging`, including the name of the individual module. This is done to avoid loading spaCy in contexts where we don't need it (importing from `utils` would execute `__init__.py`, which would in turn load every module it re-exports). We could consider moving `configure_logging()` and `get_config()` into a `generic_utils.py` that is loaded by `__init__.py`, which would allow importing from `utils`; this seems unlikely to be very useful and would cause more maintenance overhead, as we'd need to update `__init__.py` whenever we added more generic functions.

36 automated tests should pass: `docker compose exec ftva_data python -m unittest`

[DLSR-371]: https://uclalibrary.atlassian.net/browse/DLSR-371?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


utils/__init__.py
```diff
@@ -0,0 +1,11 @@
+"""
```
akohler, line 1: The context in your PR description was very helpful.  Let's make it permanent by including relevant parts of it here.  Some variation on this info should be enough:

> For this repo, imports must be written like from utils.filemaker_utils import configure_logging, including the name of the individual module. This is done to avoid loading spaCy in contexts where we don't need it (importing from utils would execute __init__.py, which would in turn load every module it re-exports)

akohler (approved):
@ztucker4 This is great.  All changes look correct and all tests still pass.  

Once the context in `utils.__init__.py` is updated (see my other comment), I'll merge.

Thanks --Andy

akohler (approved):
@ztucker4 Thanks for the documentation, looks great to me - merging.

PR #57, opened 2026-06-18
DLSR-378: Refactor `generate_metadata.py` to accept batch number

henry-r18:
Implements [DLSR-378](https://uclalibrary.atlassian.net/browse/DLSR-378)

### Acceptance criteria
- [X] Modify `generate_metadata.py` to accept a `batch_number` argument
- [X] Retrieve data from Digital Data application by `batch_number` instead of using `input_file`
- [X] Set asset/track relationships using data from DD app
- [X] Remove `input_file` support

### Description
**Update:** This PR is now intended to work with [recent changes](https://github.com/UCLALibrary/ftva-etl/pull/35) to the `ftva_etl` package. [See comment](https://github.com/UCLALibrary/ftva-mams-data/pull/57#issuecomment-4804494317) below for more info.

<s>Refactors `generate_metadata.py` to fetch a specified batch of records from the Digital Data app using the new filtering support. Much of the logic remains in place, with some functions renamed for clarity. The `input_file` argument and its related functions are removed.

The most significant change to logic relates to the `record_type` and `match_asset` properties on the output metadata. Previously, these were set via the input file; if a record was a track, it would have its `match_asset` field set to the UUID of its parent asset in the input file. Now, these relationships are derived from the `incoming_relationships` field on the records fetched from Digital Data. The idea is, while looping through the Digital Data records, we default to treating everything as an asset. Then, if we find a record with an incoming relationship of type `isTrackOf`, then we set it to `record_type=track` and set its `match_asset` accordingly. **Notably, this takes the first `isTrackOf` relationship from Digital Data; a single DD record could be a track of more than one asset, but that's not handled here.**

`ftva-etl` is also bumped to `v0.4.0` in `requirements.txt`.</s>

### Testing
Using batch `3c` as a reference, I compared the original JSON output to a newly generated output. The original output file is named `dd_records_ingest_03c_2026-04-09.json`, and I sent it to Christina on [Slack](https://uclalibrary.slack.com/archives/C0893DA0TL0/p1775756564715239?thread_ts=1775671069.228839&cid=C0893DA0TL0). Let me know if it's inaccessible, and I'll resend it. After regenerating batch `3c`, I used the scratch program below to make sure the outputs were sorted in the same way, allowing for a clean diff.

To replicate, run `python generate_metadata.py -c config_dev.toml -b 3c` against the production Digital Data instance, then run:

```python
import json

data_a = json.load(open("{{PATH_TO_ORIGINAL_OUTPUT_FILE}}"))  # change this
data_b = json.load(open("{{PATH_TO_NEW_OUTPUT_FILE}}"))  # change this

# Sort the assets by UUID
data_a["media"]["assets"].sort(key=lambda x: x["uuid"])
data_b["media"]["assets"].sort(key=lambda x: x["uuid"])

# Now make sure the keys are in the same order
data_a["media"]["assets"] = [
    {k: asset[k] for k in sorted(asset)} for asset in data_a["media"]["assets"]
]
data_b["media"]["assets"] = [
    {k: asset[k] for k in sorted(asset)} for asset in data_b["media"]["assets"]
]

json.dump(data_a, open("output/output_a_ordered.json", "w"), indent=4)
json.dump(data_b, open("output/output_b_ordered.json", "w"), indent=4)
```

Perhaps there's a way to do this using `jq`, but I don't know `jq` very well. Diff the resulting outputs however you prefer; there should be one small difference in a date value, but that must have changed after the JSON was generated originally.

[DLSR-378]: https://uclalibrary.atlassian.net/browse/DLSR-378?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/generate_metadata.py
+++ b/generate_metadata.py
@@ -1,4 +1,3 @@
-import csv
 import json
 import argparse
 import tomllib
@@ -13,7 +12,6 @@ from ftva_etl import (
 )
 from ftva_etl.metadata.utils import filter_by_inventory_number_and_library
 from requests.exceptions import HTTPError
-from warnings import deprecated
 from pymarc import Record
 
 
@@ -51,19 +49,22 @@ def configure_logging(console_logging: bool = True) -> None:
 def _get_arguments() -> argparse.Namespace:
     """Parse command line arguments.
 
-    :return: Parsed arguments for config_file, input_file, and
-    output_file as a Namespace object."""
-    parser = argparse.ArgumentParser(description="Process metadata for MAMS ingestion.")
+    :return: Parsed arguments for the program."""
+    parser = argparse.ArgumentParser(
+        description="Prepare JSON metadata for MAMS ingestion."
+    )
     parser.add_argument(
+        "-c",
         "--config_file",
         help="Path to configuration file with API credentials.",
         required=True,
     )
     parser.add_argument(
-        "--input_file",
+        "-b",
+        "--batch_number",
         type=str,
-        required=True,
-        help="Path to the input CSV file containing records to be processed.",
+        required=False,
+        help="Alphanumeric batch number to fetch records from Digital Data.",
     )
     parser.add_argument(
         "--output_dir",
@@ -71,13 +72,6 @@ def _get_arguments() -> argparse.Namespace:
         default="output/",
         required=False,
         help="Path to the output directory where JSON files will be saved. Defaults to 'output/'.",
-    )
-    parser.add_argument(
-        "--split_dpx_audio",
-        action="store_true",
-        required=False,
-        help="If specified, split output JSON into DPX, DPX Audio, and Non-DPX files.",
-        deprecated=True,
     )
     parser.add_argument(
         "--disable_console_logging",
@@ -97,16 +91,6 @@ def _get_config(config_file_name: str) -> dict:
     with open(config_file_name, "rb") as f:
         config = tomllib.load(f)
     return config
-
-
-def _read_input_file(file_path: str) -> list[dict]:
-    """Read the input CSV file and return a list of dictionaries.
-
-    :param file_path: Path to the input CSV file.
-    :return: List of dictionaries representing each row in the CSV file."""
-    with open(file_path, mode="r", encoding="utf-8-sig") as file:
-        reader = csv.DictReader(file)
-        return [row for row in reader]
 
 
 def _write_output_file(output_file: str | Path, data: dict | list[dict]) -> None:
@@ -138,6 +122,34 @@ def _initialize_clients(
             config["digital_data"]["url"],
         ),
     )
+
+
+def _get_records_by_batch_number(
+    batch_number: str,
+    digital_data_client: DigitalDataClient,
+) -> list[dict]:
+    """Get records filtered by batch number from the Digital Data application.
+
+    :param batch_number: The alphanumeric batch number to filter by.
+    :param digital_data_client: The DigitalDataClient instance to use to get the records.
+    :return: A list of records.
+    """
+    all_records: list[dict] = []
+    offset = 0
+    while True:
+        response = digital_data_client.get_records(
+            query=batch_number,
+            fields=["batch_number"],
+            offset=offset,
+        )
+        records = response.get("records", [])
+        all_records.extend(records)
+        offset += len(records)
+        # Break if pagination reaches end of records,
+        # or if a problem causes API to return no records before then.
+        if offset >= response.get("total_records", 0) or not records:
+            break
+    return all_records
 
 
 def _get_alma_bib_record(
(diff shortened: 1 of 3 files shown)


generate_metadata.py
```diff
@@ -51,19 +49,22 @@ def configure_logging(console_logging: bool = True) -> None:
...
         required=True,
     )
     parser.add_argument(
-        "--input_file",
+        "-b",
+        "--batch_number",
         type=str,
-        required=True,
-        help="Path to the input CSV file containing records to be processed.",
+        required=False,
```
akohler: This needs to be required.  The `main` routine expects a value, and doesn't abort if one is not provided:
```
2026-06-22 12:46:02,230 INFO: Fetching records for batch number None from Digital Data...
```

generate_metadata.py
```diff
@@ -176,78 +188,42 @@ def _get_alma_bib_record(
...
+        metadata_record["record_type"] = "track"
+        metadata_record["match_asset"] = track_relationship.get("source_uuid", "")
+    return metadata_record
 
 
-def _update_match_record_type(record_with_match: dict, matched_record: dict) -> dict:
-    """Check if the match_asset relationship is valid between two records. If valid,
-    set record_type to 'track' on the record_with_match. If not valid, set record_type
-    to 'asset'.
+def _compare_inventory_numbers(record_with_match: dict, matched_record: dict) -> dict:
```
akohler: This method needs a better name.  `_compare_inventory_numbers()` makes me think I'll get a Boolean response depending on whether they match.  Instead, this method sets `record_type`, depending on the result of a comparison.

I also do not understand the distinction between the `record_type` setting here and the `record_type` setting in `_set_record_type_and_match_asset()`.  Is this an override of a previous decision made in `_set_record_type_and_match_asset()`?

generate_metadata.py
```diff
@@ -176,78 +188,42 @@ def _get_alma_bib_record(
...
+            for relationship in incoming_relationships
+            if relationship.get("relationship_type", "") == "isTrackOf"
+        ),
+        None,
+    )
+    # If a track relationship is found, set `record_type=track`
+    # and `match_asset` to the source UUID for the relationship.
+    if track_relationship:
+        metadata_record["record_type"] = "track"
+        metadata_record["match_asset"] = track_relationship.get("source_uuid", "")
```
akohler: The default value in this `get()` doesn't seem appropriate.  If the DD record says it's a track of something, but doesn't have a `source_uuid`, that should be an error, or at least something should be logged - there's a data problem.

generate_metadata.py
```diff
@@ -431,45 +378,32 @@ def main() -> None:
...
-        output_dict = {"media": {"assets": metadata_records}}
-        _write_output_file(
-            Path(args.output_dir, f"{output_filename_stem}.json"), output_dict
-        )
-        logger.info(f"Output JSON file saved to '{args.output_dir}'")
+    metadata_records = _process_digital_data_records(
+        digital_data_records, alma_sru_client, filemaker_client
+    )
+
+    metadata_records = _validate_match_asset_relationships(metadata_records)
```
akohler: I think this should be called from within `_process_digital_data_records()`.  Once the list of metadata records is created, this is a mandatory part of finishing that list, so calling it there instead of from `main()` makes that coupling clearer.

generate_metadata.py
```diff
@@ -287,129 +263,100 @@ def _update_match_record_type(record_with_match: dict, matched_record: dict) ->
...
+            # Compare the inventory numbers of the related records
+            # and overwrite the `record_type` value if necessary
+            record = _compare_inventory_numbers(record, matched_record)
     return metadata_records
 
 
-@deprecated("Metadata records no longer need to be split into separate JSON outputs.")
-def _split_dpx_records(metadata_records: list[dict]) -> dict[str, list[dict]]:
-    """Split metadata records into DPX, DPX Audio, and Non-DPX categories.
+def _process_digital_data_records(
```
akohler: At this point, a better name for this method would be `create_metadata_records()`.  It does process DD records, but the result is metadata records.

generate_metadata.py
```diff
@@ -287,129 +263,100 @@ def _update_match_record_type(record_with_match: dict, matched_record: dict) ->
         return record_with_match
 
 
-def _set_record_type(metadata_records: list[dict]) -> list[dict]:
-    """Set the `record_type` for each metadata record based on its `match_asset` value.
+def _validate_match_asset_relationships(metadata_records: list[dict]) -> list[dict]:
```
akohler: I think this method name could be clearer.  Validation is more akin to data checking - is this good or not?  It doesn't reject invalid records, it updates them.  That mixture of purpose, along with the naming of `_compare_inventory_numbers()`, makes it hard to understand what these methods are doing and how responsibilities are divided among them.

akohler (changes requested):
@henry-r18 I think some of the refactoring to handle relationships makes the code less clear.  Maybe there's a parallel with my review: confusion around individual pieces of code dealing with relationships, which might be clearer looking at this as a whole, like this summary will (try to) do.

This snippet from `_process_digital_data_records()` may be the key:

```
        # Now extend metadata record with `record_type`,
        # and `match_asset` for tracks, if applicable.
        metadata_record = _set_record_type_and_match_asset(
            digital_data_record, metadata_record
        )
```

Rather than doing anything with record type & asset here, while iterating through the records, maybe defer that to the end, when you've got the full set of "metadata records".  If decisions can't be finalized until the set of records is available, make those decisions together, at the end, when you've got that set.

1. Iterate over DD records, pulling Alma & FM data for each to create metadata records.
2. Iterate over metadata records, setting record type & asset/track info based on info within the set.
3. Output finished records to file, write stats etc.

Happy to talk about it more directly if that's helpful.
Thanks --Andy

generate_metadata.py
```diff
@@ -176,240 +188,121 @@ def _get_alma_bib_record(
     return bib_record
 
 
-def _process_input_data(
-    input_data: list[dict],
+def _get_filemaker_record(
+    inventory_number: str,
+    filemaker_client: FilemakerClient,
+) -> FM_Record | None:
+    """Get the FileMaker record for the provided inventory number,
```
akohler: Clarify that this returns only the first matching record.

generate_metadata.py
```diff
@@ -140,10 +126,38 @@ def _initialize_clients(
...
+            break
+    return all_records
+
+
+def _get_alma_bib_record_with_possible_suffix(
     inv_no_stem: str,
     alma_sru_client: AlmaSRUClient,
-) -> Record | None:
+) -> Pymarc_Record | None:
     """Get the Alma bib record for the provided inventory number,
```
akohler: Similar to `_get_filemaker_record()`, note that this returns only the first matching record.  This method also has a comment at the slice, which is good.

akohler (changes requested):
@henry-r18 This is a big improvement.  With code moved to the `ftva_etl` package and the changes to how the batch is validated, the program flow is much clearer.

I agree with your decision to make the batch fail if any asset relationship fails.  That fits nicely with the hope that this legacy batch processing can be incorporated into the Digital Data application itself, removing the user <-> developer feedback cycle altogether.

I flagged 2 very small changes to clarify that the FM and Alma record routines retrieve only the first matching record... which was existing behavior, just surfaced via your refactoring.

Once those are made, I'll merge this.

Thanks --Andy

akohler (approved):
@henry-r18 👍 All good now, merged.  Thanks --Andy

PR #60, opened 2026-07-08
DLSR-451 and DLSR-452: Add metadata generation script for NDM and refactor code

henry-r18:
Implements [DLSR-451](https://uclalibrary.atlassian.net/browse/DLSR-451) and [DLSR-452](https://uclalibrary.atlassian.net/browse/DLSR-452) (split off from [DLSR-440](https://uclalibrary.atlassian.net/browse/DLSR-440))

## Acceptance criteria
- [X] DLSR-451: Create a script similar to `generate_metadata.py` which:
  - Reads batch information from a CSV file
  - Retrieves the relevant data needed for JSON per the [NDM ETL specs](https://uclalibrary.atlassian.net/wiki/x/BIBkb)
    - Filemaker item records via `New Digital DMIU` layout
    - Filemaker inventory records via `NEW DIGITAL_API` layout
    - Alma bib records from Alma
- [X] DLSR-452: Copy shared code from existing `generate_metadata.py` to modules for reuse
  - Refactoring the legacy script should be a separate ticket, if needed
#### To-do
- Assemble JSON using new function in `ftva_etl` package and write to file

## Description
The changes included in this PR would:
- Add `generate_metadata_new_media.py` to generate MAMS JSON metadata for new digital media (NDM) records from a CSV batch file, FileMaker, and Alma
- Rename `generate_metadata.py` to `generate_metadata_legacy_media.py` and refactor it to use shared utilities
  - The behavior for the legacy workflow is unchanged
- Extract shared logic into `utils/alma_utils.py` (Alma bib lookup) and `utils/generate_metadata_utils.py` (general helpers and batch validation code)
- Update tests:
  - Rename `tests/test_generate_metadata.py` to `tests/test_generate_metadata_utils.py`
  - Improve the way tests for the `validate_match_asset_relationships` helper handle logging assertions

I liked the way @ztucker4 organized `filemaker_validation_report.py`, so I mimicked that style with the section comment blocks.

In `generate_metadata_new_media.py`, `_get_metadata_records()` calls a function imported from `ftva_etl` named `get_mams_metadata_ndm`. This doesn't exist yet; DLSR-447 and 448 cover the work for adding it to the `ftva_etl` package. However, it does outline the data that the `ftva_etl` package should expect:
- `item_record`: An `fmrest` Record representing a record from the `Digital Media_Item Unit` table, accessible via the `New Digital DMIU` layout
- `inventory_record`: An `fmrest` Record representing a record from the `Inventory for Labeling` table, accessible via the `NEW DIGITAL_API`
- `alma_bib_record`: A `pymarc` Record if a bib record is found matching the `inventory_no` from the FM `inventory_record`, or `None` otherwise
- `match_asset`: A UUID string from the input CSV file, or `None` if not provided

I figured it was good to know what the input should be for the new `ftva_etl` work should be before getting started on it, hence why I'm submitting this PR first.

## Testing
I didn't add any new unit tests, but I did adjust the existing tests to work with the utility functions that were moved around.

[DLSR-451]: https://uclalibrary.atlassian.net/browse/DLSR-451?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DLSR-452]: https://uclalibrary.atlassian.net/browse/DLSR-452?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DLSR-440]: https://uclalibrary.atlassian.net/browse/DLSR-440?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


generate_metadata_new_media.py
```diff
@@ -0,0 +1,234 @@
...
+    """Get all item unit records indexed by UUID
+    from the "New Digital DMIU" layout in Filemaker.
+
+    :param config: Config dict with Filemaker API credentials.
+    :return: A dict of item unit records indexed by UUID.
+    """
+
+    # Item unit records are stored in the "Digital Media Item Unit" table,
+    # accessed via the "New Digital DMIU" layout
+    fm_item_unit_layout = "New Digital DMIU"
```
akohler, line 59: I wasn't aware of this layout, and don't see it in Confluence or Jira.  It does work :) - did you discover it via the FM client or did I just miss it elsewhere?

generate_metadata_new_media.py
```diff
@@ -0,0 +1,234 @@
...
+
+# Module-level logger used throughout this module.
+# Handlers are configured explicitly via `_configure_logging`.
+LOGGER = logging.getLogger(Path(__file__).stem)
+
+
+# ---------------------------------------------------------------------------
+# Data fetching
+# ---------------------------------------------------------------------------
+def _get_fm_inventory_records_indexed_by_inventory_id(
```
akohler: I had expected this script to retrieve inventory records individually, by `inventory_id` (supplied by the item record's `inventory_id_fk`).

Retrieving all digital inventory records instead does make  sense for now, given that the number of these records is very low (I see just 19 in the production db at the moment).  That will (had better!) increase over time, meaning the time to retrieve all digital inventory records will increase, probably linearly.

Given that batches are likely to be small (on the order of 50 items), the time to retrieve up to that many inventory records probably will be less than the time to retrieve all inventory records before too long.

I know, and agree with, the arguments against [premature optimization](https://en.wikiquote.org/wiki/Donald_Knuth#Computer_Programming_as_an_Art_(1974)), so I'm just noting this at the moment.  A record-by-record implementation, with basic caching in the likely case that multiple items in a batch go with the same inventory record, should be easy to do if/when warranted.

generate_metadata_new_media.py
```diff
@@ -0,0 +1,234 @@
...
+        config["filemaker"]["password"],
+        layout=fm_inventory_layout,
+    )
+    # `digital_record` field necessary to filter for digital inventory records
+    query = [{"digital_record": "1"}]
+    fm_inventory_records = fm_client.find_all_records(query=query)
+    return {record["inventory_id"]: record for record in fm_inventory_records}
+
+
+def _get_fm_item_records_indexed_by_uuid(config: dict) -> dict[str, Record]:
```
akohler: Similar to my comment on `_get_fm_inventory_records_indexed_by_inventory_id()`, I'm a little surprised by retrieving all item records instead of just the ones in the current batch.  There are more item records, currently 117, suggesting about a 6-1 ratio with digital inventory records.

Unlike inventory records, the item records should be a 1-1 match with the size of each batch.  With roughly 50 records/batch, I don't see the value of retrieving all items and filtering locally, especially as the number of items in the db will grow pretty quickly to 1000+.

I'm open to arguments for the current implementation, but I'm flagging this for change unless I'm persuaded :)

generate_metadata_new_media.py
```diff
@@ -0,0 +1,234 @@
...
+            item_record, inventory_record, alma_bib_record, match_asset
+        )
+        metadata_records.append(metadata_record)
+    return metadata_records
+
+
+# ---------------------------------------------------------------------------
+# CLI arguments, logging, and input file loading
+# ---------------------------------------------------------------------------
+def _read_input_file(input_file: str | Path) -> list[dict]:
```
akohler: This seems like a candidate for the `generate_metadata_utils` module, similar to `write_output_file`.  There's nothing specific to the current script about this method, other than the commonly-used `list[dict]` return type.

generate_metadata_new_media.py
```diff
@@ -0,0 +1,234 @@
...
+    """Read the input file and return the data as a list of dictionaries.
+
+    :param input_file: Path to the input CSV file.
+    :return: A list of dictionaries."""
+    with open(input_file, "r", encoding="utf-8") as file:
+        return [row for row in csv.DictReader(file)]
+
+
+def _configure_logging(console_logging: bool = True) -> None:
```
akohler: This `_configure_logging()` looks identical to the one still in the legacy script, so seems worth moving to the utils module.  In fact, there's one in `utils.filemaker_utils.py` now, exported via `utils.__init__.py()`.

akohler (changes requested):
@henry-r18 I like this approach.  I think the general flow of the new program is correct, and this provides the scaffolding to be filled in as details are added to the ETL package.

Tests all pass, and the updated legacy script regenerated batch 3C JSON correctly - except for the date bug.  Please update `requirements.txt` to pull in the [fixed version](https://github.com/UCLALibrary/ftva-etl/releases/tag/v0.5.1).

The refactoring is good, showing the evolution of this collection of work.  I expected the legacy refactoring to be done in a separate ticket, but since it has started :), I called out a couple of places where it should also be done, since duplicate code has been added to the new script.

I also added a non-review comment about the implementation of `_get_fm_inventory_records_indexed_by_inventory_id()`.  It's OK for now but I'm flagging possible scaling issues as the number of digital inventory records grows.  If there's a strong reason for the current implementation, it's probably worth recording it in the method documentation.  I'm OK with keeping it as implemented for now, also OK if you decide to change it now.

Thanks --Andy

generate_metadata_new_media.py
henry-r18: @akohler I found this layout when poking around via the API. I noticed there's a `get_layouts()` method on the `fmrest.Server` class (i.e. our `FilemakerClient._fms`). Here's a gist that lists out the available layouts:

```
>>> from pprint import pprint
>>> from ftva_etl import FilemakerClient
>>> fm_client = FilemakerClient(USERNAME, PASSWORD)
>>> layouts = fm_client._fms.get_layouts()
>>> pprint(layouts)
```

I also didn't see it documented anywhere. However, the other layouts--`NEW DIGITAL_API` and `NEW DIGITAL STORAGE_API`--don't appear to return the right view of the data for this script. `NEW DIGITAL_API` does allow us to see related records from the "Digital Media Item Unit" table, but it only "joins" certain fields and notably not the `UUID` field on the DMIU records.

I did consider bringing it up in Slack and asking that the portal/view be adjusted to include the `UUID` field on the DMIU records. But since we can go from the item records to the inventory records via the `inventory_id_fk` foreign key, I didn't think it was necessary to bug Salik with it...

akohler: @henry-r18 Makes sense, and since it works, totally appropriate.  I thought Salik had to enable layouts for API access, or even copy them to create API-specific layouts, thus my surprise.  Thanks.

generate_metadata_new_media.py
```diff
@@ -0,0 +1,209 @@
...
+def _get_inventory_record_by_inventory_id(config: dict, inventory_id: int) -> Record:
+    """Get an inventory record by inventory ID from the "NEW DIGITAL_API" layout in Filemaker.
+
+    :param config: Config dict with Filemaker API credentials.
+    :param inventory_id: Inventory ID of the inventory record to get.
+    :return: The inventory record.
+    :raises SystemExit: if the inventory record is not found.
+    """
+    fm_inventory_layout = "NEW DIGITAL_API"
+    fm_client = FilemakerClient(
```
akohler: Instead of initializing a client for every query (which made more sense when this was fetching all records, via 1 query), initialize it in `_get_metadata_records()` along with the Alma client.  Then pass the FM client into the methods where FM data retrieval is done.

Layout can be a parameter to the "find" methods via their `**kwargs`.  This is buried several layers up; see [`fmrest.Server.find` docs](https://github.com/davidhamann/python-fmrest/blob/master/fmrest/server.py#L613), specifically `request_layout` and `response_layout`.

generate_metadata_new_media.py
```diff
@@ -0,0 +1,209 @@
...
+def _get_item_record_by_uuid(config: dict, uuid: str) -> Record:
+    """Get an item record by UUID from the "New Digital DMIU" layout in Filemaker.
+
+    :param config: Config dict with Filemaker API credentials.
+    :param uuid: UUID of the item record to get.
+    :return: The item record.
+    :raises SystemExit: if the item record is not found.
+    """
+    fm_item_unit_layout = "New Digital DMIU"
+    fm_client = FilemakerClient(
```
akohler: Same as for `_get_inventory_record_by_inventory_id()`, pass in an initialized client and then pass the layout along with the query.

akohler (changes requested):
@henry-r18 The refactoring looks very good, thank you.  Just two more changes, to avoid re-initializing the FM client for every search, and I think this will be ready to go.

akohler (approved):
@henry-r18 All looks good to me now.  I will assume the right records are found via the `request_layout` parameter :) - if not, something to dig deeper into when the rest of the wiring is done.

Merging, thank you!

PR #61, opened 2026-07-09
DLSR-439: schedule FM scripts

ztucker4:
Implements/relates to [DLSR-439](https://uclalibrary.atlassian.net/browse/DLSR-439)

This PR includes a new shell script, `run_filemaker_scripts.sh`, intended for use on the support server (`p-u-exlsupport01`). The script first runs `filemaker_batch_update.py` for all currently supported fields (`production_type`, `Language`, `director`, `release_broadcast_year`, and `record_date`). Then `filemaker_validation_report.py` is run against all supported layouts (`InventoryForLabeling_API`, `NEW DIGITAL_API`, and `NEW DIGITAL STORAGE_API`) for records updated in the last two weeks. Output CSVs from the validation report are compressed if they are larger than 1MB (as was the case for the `InventoryForLabeling_API` layout, run this morning). Finally, log files and potentially-compressed CSVs are sent as email attachments to Staci and Amanda, cc Andy Kohler. 

This PR also includes a change to `requirements.txt` to pin NumPy to version 2.3.5. This is the last version of NumPy released before the X86_V2 microarchitecture became required (see [NumPy 2.4.0 release notes](https://numpy.org/devdocs/release/2.4.0-notes.html#modulate-dispatched-x86-cpu-features)). This release is compatible with Python 3.13, and I have tested `generate_metadata.py` against a few batches of records - downgrading NumPy has not affected the output JSON records. 

On the server, I have installed uv for the `exlsupport` user and created a venv in this repo's directory using `uv pip install -r requirements.txt`. I also created a `.python_version` file setting uv's python version to 3.13. 

Once the code is merged and pulled onto the server, I intend to add this line to the server's crontab:
```
0 9 * * 1 [ $(( $(date +\%V) \% 2 )) -eq 1 ] && /home/exlsupport/ftva-mams-data/run_filemaker_scripts.sh
```

This will run the script at 9AM on Mondays every other week (determined by parity of the ISO week number). Using odd weeks would start the process this coming Monday, 7/13/26.

For testing, I have a version of the script on `p-u-exlsupport01` (`run_filemaker_scripts_zl_test.sh`) that only executes the reporting script and sends email to only me, plus a commented-out line in the crontab for a scheduling a one-off run.

[DLSR-439]: https://uclalibrary.atlassian.net/browse/DLSR-439?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


requirements.txt
```diff
@@ -13,3 +13,4 @@ retry==0.9.2
 # For date parsing
 edtf==5.0.1
 dateparser==1.4.0
+numpy==2.3.5
```
akohler, line 18: Add a comment to document that we're pinning `numpy` to this version deliberately.  Otherwise, developers are likely to see this is old and update it.

akohler (changes requested):
@ztucker4 This looks very good to me.  I tested on the server by copying your test script, modifying the output files to `/tmp/AK_*`, and ran it on the command line.  I also ran it via cron, with no errors.

Other than the review comment requesting documentation in `requirements.txt`, I have a few general notes which don't fit well in line-by-line code review:
1. Add start and end times to the output captured in script log files.  My impression is that it took 20+ minutes to run the test script, doing just validation, but I can't tell that from the validation log file.  The Python log files do have timestamps on messages but I want an overall understanding of when the full process ran and how long it took.
2. CSV files for layouts with no validation errors are created and emailed, with a single row for the report headers.  This is confusing for users.  You could either modify the validation program itself to not write header-only CSVs, or modify the script to not email them.  The former seems cleaner.
3. The compressed file is `.gz`, which is fine for us (and may be needed, as I think Outlook blocks "dangerous" `.zip` files because Windows does stupid stuff based on file extensions...).  We'll need to confirm that the FTVA recipients can receive and unpack `.gz` files successfully.
4. 163K violations in the legacy layout in 2 weeks... this may indicate a problem with the validation script itself.  That's outside the scope of this PR but should be reviewed.

Thanks --Andy

akohler (approved):
@ztucker4 All looks good to me.  I don't have time today to re-run this locally, but all the new changes look completely correct to me.

Merging now.  Please go ahead and deploy to the server and get this scheduled however the staff decided.

Thanks --Andy

PR #62, opened 2026-07-21
DLSR-463: Make inventory number check optional

henry-r18:
Implements [DLSR-463](https://uclalibrary.atlassian.net/browse/DLSR-463)

## Acceptance criteria
- [X] The check for matching inventory numbers between tracks and assets in `utils.generate_metadata_utils.validate_match_asset_relationships` is made optional
- [X] Consumers of the function are updated accordingly

## Description
Per discussion [on Slack](https://uclalibrary.slack.com/archives/C0893DA0TL0/p1783717736894839), the validation rules for legacy and NDM metadata should be different, with legacy metadata requiring that: 1) the value of the match_asset field on a track actually refers to a UUID of an item in the batch; and 2) the inventory numbers between the related items match. However, NDM metadata only needs to satisfy the first check.

This PR makes the inventory number comparison optional in the validation utility that is called in both the legacy and new media metadata generation scripts.

The `ftva-etl` package is also updated to `v0.6.0`

## Testing
1 new unit test is added to cover the validation utility with `check_inventory_numbers = False`, and a bug causing a failing test is fixed. There are 37 tests all passing: `python -m unittest`

Using the "NDM_Media to Ingest into MAMs" Google Sheet as input (can share separately, as the link is publicly accessible), running `python generate_metadata_new_media.py -c {TOML_CONFIG} -i {PATH_TO_SPREADSHEET_CSV}` should return the following error, resulting from an underlying data quality issue in Filemaker. The `creation_date` field in FM is not allowed to be anything other than a date, so the value `?` is raising this error, which is the desired behavior.

**Expected error**
```
  File "/home/ftva_data/project/generate_metadata_new_media.py", line 116, in _get_metadata_records
    metadata_record = get_mams_metadata_ndm(
        item_record, inventory_record, alma_bib_record, match_asset, nlp_model
    )
  File "/home/ftva_data/.local/lib/python3.13/site-packages/ftva_etl/metadata/mams_metadata_ndm.py", line 74, in get_mams_metadata_ndm
    "creation_date": get_creation_date(fm_item_record),
                     ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^
  File "/home/ftva_data/.local/lib/python3.13/site-packages/ftva_etl/metadata/filemaker.py", line 141, in get_creation_date
    raise ValueError(message) from e
ValueError: Failed to parse creation date for record 99: '?' cannot be parsed to a date
```

Using a test copy of the full batch CSV with only the first 5 rows should result in a successful output.

[DLSR-463]: https://uclalibrary.atlassian.net/browse/DLSR-463?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/generate_metadata_legacy_media.py
+++ b/generate_metadata_legacy_media.py
@@ -209,13 +209,14 @@ def main() -> None:
         digital_data_records, alma_sru_client, filemaker_client
     )
 
-    # If match_asset relationships are invalid, log an error and exit
-    if not gm_utils.validate_match_asset_relationships(
-        metadata_records=metadata_records,
-        logger=LOGGER,
-    ):
+    # If there are any validation problems, log them and exit without writing output file
+    validation_problems = gm_utils.validate_match_asset_relationships(metadata_records)
+    if validation_problems:
+        for problem in validation_problems:
+            LOGGER.error(problem)
         LOGGER.error(
-            "Invalid match_asset relationships found in metadata records. Review logs for details."
+            "Problems found with match_asset relationships. "
+            "Please fix the issues and try again."
         )
         return
 
--- a/generate_metadata_legacy_media.py
+++ b/generate_metadata_legacy_media.py
@@ -210,7 +210,10 @@ def main() -> None:
     )
 
     # If match_asset relationships are invalid, log an error and exit
-    if not gm_utils.validate_match_asset_relationships(metadata_records):
+    if not gm_utils.validate_match_asset_relationships(
+        metadata_records=metadata_records,
+        logger=LOGGER,
+    ):
         LOGGER.error(
             "Invalid match_asset relationships found in metadata records. Review logs for details."
         )
--- a/generate_metadata_new_media.py
+++ b/generate_metadata_new_media.py
@@ -180,7 +180,12 @@ def main() -> None:
     metadata_records = _get_metadata_records(config, input_data)
 
     # If match_asset relationships are invalid, log an error and exit
-    if not gm_utils.validate_match_asset_relationships(metadata_records):
+    # Inventory numbers are expected to differ for NDM, so we disable inventory number validation.
+    if not gm_utils.validate_match_asset_relationships(
+        metadata_records=metadata_records,
+        check_inventory_numbers=False,
+        logger=LOGGER,
+    ):
         LOGGER.error(
             "Invalid match_asset relationships found in metadata records. Review logs for details."
         )
--- a/generate_metadata_new_media.py
+++ b/generate_metadata_new_media.py
@@ -179,15 +179,17 @@ def main() -> None:
 
     metadata_records = _get_metadata_records(config, input_data)
 
-    # If match_asset relationships are invalid, log an error and exit
+    # If there are any validation problems, log them and exit without writing output file.
     # Inventory numbers are expected to differ for NDM, so we disable inventory number validation.
-    if not gm_utils.validate_match_asset_relationships(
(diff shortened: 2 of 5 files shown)


utils/generate_metadata_utils.py
```diff
@@ -39,18 +45,20 @@ def validate_match_asset_relationships(
...
-        if record_inv != matched_record_inv:
-            _logger.error(
-                f"Inventory numbers do not match for match_asset relationship "
-                f"{record['record_type']} {record['uuid']}: '{record_inv}', "
-                f"{matched_record['record_type']} {matched_record['uuid']}: '{matched_record_inv}'"
-            )
-            return False
+        if check_inventory_numbers:
+            # Now check inventory numbers, using the first inv no for each record
+            record_inv = (record.get("inventory_numbers") or [None])[0]
```
akohler: I prefer the simple form of these defaults, without the boolean and the extra parens, since `get()` can return any default needed:
```
record_inv = record.get("inventory_numbers", [None])[0]
matched_record_inv = matched_record.get("inventory_numbers", [None])[0]
```

generate_metadata_legacy_media.py
```diff
@@ -210,7 +210,10 @@ def main() -> None:
     )
 
     # If match_asset relationships are invalid, log an error and exit
-    if not gm_utils.validate_match_asset_relationships(metadata_records):
+    if not gm_utils.validate_match_asset_relationships(
+        metadata_records=metadata_records,
+        logger=LOGGER,
```
akohler: Why does `LOGGER` need to be passed into this method now, when it didn't before?  `validate_match_asset_relationships()` did logging before without this being passed in.  Not necessarily wrong now, but is this change needed?

generate_metadata_new_media.py
```diff
@@ -180,7 +180,12 @@ def main() -> None:
     metadata_records = _get_metadata_records(config, input_data)
 
     # If match_asset relationships are invalid, log an error and exit
-    if not gm_utils.validate_match_asset_relationships(metadata_records):
+    # Inventory numbers are expected to differ for NDM, so we disable inventory number validation.
+    if not gm_utils.validate_match_asset_relationships(
+        metadata_records=metadata_records,
+        check_inventory_numbers=False,
+        logger=LOGGER,
```
akohler: Same note on the `LOGGER` passed in here as with the legacy script change.  Since these scripts use module-level logging, I'm suspicious when I see loggers passed around - but I have lost track of when it's still necessary, so feel free to correct me!

akohler (changes requested):
@henry-r18 I think this is functionally correct. The tests are good, code and comments are clear.  

I'm requesting a small change as I prefer using explicit defaults: `get(some_list, [None])` over the generally less readable `(get(some_list) or [None])`. 

I'm also questioning the module-level `LOGGER` being passed into methods which have been imported into the module.  I'm pretty sure that's not needed - that's why it's defined at the module level - but I admit some uncertainty so am happy to being convinced otherwise :)

Thanks --Andy

akohler (approved):
@henry-r18 I like this change.  Collecting the validation problems and letting the calling script decide how to deal with them as a whole provides flexibility while moving the responsibility to the script, which seems completely appropriate to me.  This implementation does not make using the method any harder, simplifies the signature, and removes some logging confusion.

All great, merging now. Thanks --Andy

PR #63, opened 2026-07-22
DLSR-478: Improve error handling to collect all errors before exiting

henry-r18:
## Acceptance criteria
- [X] While iterating input records in `generate_metadata_new_media._get_metadata_records`, errors raised by calls to `ftva-etl.get_mams_metadata_ndm` are logged and the script continues
- [X] The script prints a message flagging errors before exiting gracefully, rather than failing with stack trace

## Description
Improves handling of errors raised by `ftva-etl.get_mams_metadata_ndm()` by wrapping the caller in a `try...except` block, which logs and continues after anticipated errors, rather than exiting on the first such error. This has the benefit that the whole batch (i.e. every row in the input CSV file) can be seen by the script before gracefully exiting with a message to "review logs for details", as opposed to crashing with a stack trace on every individual error.

For consistency's sake, we could assemble a list of messages then print them if there are any, like in #62. But for some reason, the way it is here scans more clearly to me in this context...

## Testing
No new unit tests added; 37 existing tests still pass.

Using the "NDM_Media to Ingest into MAMs" Google Sheet as input, running `python generate_metadata_new_media.py -c {TOML_CONFIG} -i {PATH_TO_SPREADSHEET_CSV}` should complete successfully, with a valid JSON output. It seems that the FM data quality issue related to the `creation_date` field was resolved, though I'll confirm that on Slack.

### DIFF
--- a/generate_metadata_new_media.py
+++ b/generate_metadata_new_media.py
@@ -79,13 +79,16 @@ def _get_item_record_by_uuid(fm_client: FilemakerClient, uuid: str) -> Record:
 # ---------------------------------------------------------------------------
 # Metadata generation
 # ---------------------------------------------------------------------------
-def _get_metadata_records(config: dict, input_data: list[dict]) -> list[dict]:
+def _get_metadata_records(
+    config: dict, input_data: list[dict]
+) -> tuple[list[dict], bool]:
     """Get metadata records for MAMS ingest,
     using input data to fetch necessary sources from Filemaker and possibly Alma.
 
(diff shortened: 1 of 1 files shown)


akohler (approved):
@henry-r18 This is all fine with me.  Using a tuple to return data and status is OK, and this flows well enough.  The data problem in this batch seems to have been fixed - I ran the full batch, same as yesterday, and no errors were reported in the logs - but I'm sure bad data will crop up again :)

Merged, thanks --Andy

PR #66, opened 2026-07-24
DLSR-485: Fix bug related to Unicode encoding

henry-r18:
Fixes [DLSR-485](https://uclalibrary.atlassian.net/browse/DLSR-485)

## Bug description
- In the JSON output from `generate_metadata_new_media.py`, there was at least one record with an escaped Unicode character in the `title` field (i.e. `{..."title": "She Don\u2019t Fade"...}`). The character should be output normally, rather than in its escaped form.

## Fix
`generate_metadata_new_media.py` calls `utils.generate_metadata_utils.write_output_file` to handle writing the metadata dict to a JSON file. Setting `ensure_ascii=False` on `json.dump` in the writing utility resolves the issue with the escaped single right quotation mark (`\u2019`), and it should address similar issues.

If we want a more targeted approach, we could create a mapping in `ftva-etl` to replace certain Unicode characters, such as punctuation marks, with their ASCII equivalents, but I don't really think that's necessary. This is a one-line change, where as that would be more involved.

## Tests
No practical unit tests to add, but please confirm the single quote looks un-escaped by re-generating the JSON for the current batch. My IDE highlights the character and gives a message that says `The character U+2019 "’" could be confused with the ASCII character U+0060...`

[DLSR-485]: https://uclalibrary.atlassian.net/browse/DLSR-485?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/utils/generate_metadata_utils.py
+++ b/utils/generate_metadata_utils.py
@@ -92,7 +92,7 @@ def write_output_file(output_file: str | Path, data: dict | list[dict]) -> None:
     # Allows for `output_file` to be a relative path.
     output_path.parent.mkdir(parents=True, exist_ok=True)
     with open(output_path, mode="w", encoding="utf-8") as file:
-        json.dump(data, file, indent=4)
+        json.dump(data, file, indent=4, ensure_ascii=False)
 
 
 def get_config(config_file_name: str) -> dict:


akohler (approved):
@henry-r18 Good fix. Interesting that `json.dump()` defaults `ensure_ascii=True`, since JSON is [supposed to be in UTF-8](https://www.rfc-editor.org/info/rfc8259/#section-8.1)... but now we know.

I totally agree that mapping specific characters in the Filemaker data back to ASCII or something else is not the right solution.  This isn't bad data - people have been copy/pasting MSWord's "smart" quotes for years - so our job is just to output the data in UTF-8 JSON, without turning it into escape sequences.  If any characters cause problems within the MAMS for some reason, we can help staff clean up the source data, or fall back to a custom output mapping if truly necessary.

Merged, thanks --Andy

